PR #4, opened 2018-11-22
fix issue when using the webhook parameter

rauny-brandao:
Hello there @iynere and @eddiewebb,

I'm having the same issue described here: [issues/2](https://github.com/CircleCI-Public/slack-orb/issues/2). The problem happens because the env **$SLACK_WEBHOOK** is always checked, even when the parameter **webhook** is set.

What do you think about this solution? Is there a better way to fix it?

Thanks!


eddiewebb:
That's the proper way to handle this, thanks for the contribution!

Issue #3, opened 2018-11-14
Would be nice to show the branch, author, and have an opportunity to rerun


thomascmost:
This is all functionality that came with the CircleCI 1.0 Slack webhook notifier -- Are there plans to add these back in?

A success or fail should include the:
- commit hash
- commit message
- commit author
- build time (?)

and in addition to the `Visit Job` button, there should be a `Rebuild` button for failures

Thank you!

thomascmost:
Closing because these sorts of details can be configured using the custom `notify` hook, and by referencing CircleCI's docs on Built-In environment variables here: https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

Example of a custom notify hook: https://github.com/CircleCI-Public/circleci-orbs/blob/151c8edfd068c39a85534126885607026afcbafb/.circleci/config.yml#L107

See #7 for additional discussion around this.

PR #7, opened 2018-11-26
Add job name to default message on Success/Fail hooks

thomascmost:
Self-explanatory -- if a job succeeds or fails it would be good to report *which* job it was.

### DIFF
--- a/src/@orb.yml
+++ b/src/@orb.yml
@@ -98,12 +98,12 @@ commands:
                   echo "The job completed successfully"
                   echo '"fail_only" is set to "true". No Slack notification sent.'
                 else
-                  curl -X POST -H 'Content-type: application/json' --data "{ \"attachments\": [ { \"fallback\": \"A job has succeeded - $CIRCLE_BUILD_URL\", \"text\": \":tada: A job has succeeded! $SLACK_MENTIONS\", \"fields\": [ { \"title\": \"Project\", \"value\": \"$CIRCLE_PROJECT_REPONAME\", \"short\": true }, { \"title\": \"Job Number\", \"value\": \"$CIRCLE_BUILD_NUM\", \"short\": true } ], \"actions\": [ { \"type\": \"button\", \"text\": \"Visit Job\", \"url\": \"$CIRCLE_BUILD_URL\" } ], \"color\": \"#1CBF43\" } ] } " << parameters.webhook >>
+                  curl -X POST -H 'Content-type: application/json' --data "{ \"attachments\": [ { \"fallback\": \"A job has succeeded - $CIRCLE_BUILD_URL\", \"text\": \":tada: A `$CIRCLE_JOB` job has succeeded! $SLACK_MENTIONS\", \"fields\": [ { \"title\": \"Project\", \"value\": \"$CIRCLE_PROJECT_REPONAME\", \"short\": true }, { \"title\": \"Job Number\", \"value\": \"$CIRCLE_BUILD_NUM\", \"short\": true }, { \"title\": \"Commit\", \"value\": \"$CIRCLE_SHA1\", \"short\": true } ], \"actions\": [ { \"type\": \"button\", \"text\": \"Visit Job\", \"url\": \"$CIRCLE_BUILD_URL\" } ], \"color\": \"#1CBF43\" } ] } " << parameters.webhook >>
                   echo "Job completed successfully. Alert sent."
                 fi
               else
                 #If Failed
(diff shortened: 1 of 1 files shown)


src/@orb.yml
```diff
@@ -98,12 +98,12 @@ commands:
...
-                  curl -X POST -H 'Content-type: application/json' --data "{ \"attachments\": [ { \"fallback\": \"A job has succeed - $CIRCLE_BUILD_URL\", \"text\": \":tada: A job has succeeded! $SLACK_MENTIONS\", \"fields\": [ { \"title\": \"Project\", \"value\": \"$CIRCLE_PROJECT_REPONAME\", \"short\": true }, { \"title\": \"Job Number\", \"value\": \"$CIRCLE_BUILD_NUM\", \"short\": true } ], \"actions\": [ { \"type\": \"button\", \"text\": \"Visit Job\", \"url\": \"$CIRCLE_BUILD_URL\" } ], \"color\": \"#1CBF43\" } ] } " << parameters.webhook >>
+                  curl -X POST -H 'Content-type: application/json' --data "{ \"attachments\": [ { \"fallback\": \"A job has succeeded - $CIRCLE_BUILD_URL\", \"text\": \":tada: A `$CIRCLE_JOB` job has succeeded! $SLACK_MENTIONS\", \"fields\": [ { \"title\": \"Project\", \"value\": \"$CIRCLE_PROJECT_REPONAME\", \"short\": true }, { \"title\": \"Job Number\", \"value\": \"$CIRCLE_BUILD_NUM\", \"short\": true }, { \"title\": \"Commit\", \"value\": \"$CIRCLE_SHA1\", \"short\": true } ], \"actions\": [ { \"type\": \"button\", \"text\": \"Visit Job\", \"url\": \"$CIRCLE_BUILD_URL\" } ], \"color\": \"#1CBF43\" } ] } " << parameters.webhook >>
```
eddiewebb: Buttons are a nice touch Kyle added before, should we update job name their or keep generic "Visit Job"

iynere:
@eddiewebb @thomascmost i actually disagree w/ hardcoding this kind of information into the `curl` call itself. this level of customization is already easy & possible w/ the existing orb:

https://github.com/CircleCI-Public/circleci-orbs/blob/151c8edfd068c39a85534126885607026afcbafb/.circleci/config.yml#L107

while it would help a common use-case, i wouldn't want to prescribe how the notification functionality should work or what it should do at that high of a level..... thoughts ? @KyleTryon 

what i _do_ think would be beneficial, is better documentation around writing rich / contextual notification messages—i just added some of that info to the readme today (but it can probably be further built out via examples, etc):

https://github.com/CircleCI-Public/slack-orb#notify

or possibly some modifications to the orb—new parameters, etc.—that more easily support a customizable notification message

thomascmost:
Totally understand. Can we agree that adding the job name is a more sensible default though? I'll back out the commit hash.

thomascmost:
@eddiewebb Thanks for the quick response -- I slimmed down my requested change.

For future improvements, it would be great to have customization parameters in the orb (see #3) -- I recognize that you support arbitrary custom messages, but for non-power-users (like myself) it's more comfortable to work within the existing framework of fail/success hooks.

Thanks again for your attention on this

KyleTryon:
I could see the argument for adding the job name. My question would be, if it can easily be added in the message itself, should we? You can implement your own formatting in messages as well. I think it may be best to keep it as minimal as possible to allow the users to build up what they want from the notification.

There are some valid requests but I wonder if the tool already provides a way of achieving this as is.

thomascmost:
Perhaps I am misunderstanding -- I didn't realize it was possible to add a custom message to the success/fail hooks. If that is the case, there could be better documentation around that. If that is not the case (it doesn't **look** like it is) then naming WHAT job has succeeded or failed still seems like an, I would argue *obvious*, sensible default.

src/@orb.yml
thomascmost: Since the job name might be long or ugly, from a UX perspective I would argue it's better if the action button is kept consistent -- providing that the job is named above

iynere:
@thomascmost if you look at the source code for the orb, there's not much difference between the 'success/fail' and the 'notify' commands...

we could add a 'custom message' field to the 'success/fail' command, but at that point why bother having separate commands at all, you know ?

can you use the 'notify' command to achieve what you're looking for ?

iynere:
i do agree, though, that `job name` & `job number` are probably good defaults for success/fail

thomascmost:
@iynere Yes, definitely. I understand CircleCI configuration much better now than I did a few weeks ago, and I'll close #3 with a comment about the custom notify command.

At this point I mostly just think that, for the sake of someone who does not really want to concern themselves with writing custom notify hooks in their config, we should make the generic success/fail hooks at least report which job they are referring to.

iynere:
in that case, this looks good to me ! 

thanks for your patience @thomascmost !!

iynere:
what do u think ? @eddiewebb

PR #10, opened 2018-11-28
Remove Backticks (incorrectly escaping the job name)

thomascmost:
Backticks are escaping/removing the job name in the @orb.yaml default message.

Sorry about that -- was going for the slack formatting but I think this is an unsupported interaction in their API


iynere:
good catch @thomascmost

thomascmost:
Thanks for the quick update @iynere

PR #15, opened 2018-12-19
Remove $SHELL check and prefer run option

zzak:
I found this in combination with `apk add bash curl` works on alpine based images.

### DIFF
--- a/src/@orb.yml
+++ b/src/@orb.yml
@@ -37,6 +37,14 @@ commands:
         default: ""
     steps:
       - run:
+          name: Provide error if non-bash shell	
+          command: |
+            if [ $(echo $SHELL | grep -c bash) -ne 1 ]; then	
+              echo Bash not installed.	
+              exit 1	
+            fi
+             
+      - run:
           name: Slack Notification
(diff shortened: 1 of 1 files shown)


iynere:
@zzak why not just add `apk add bash curl` to the orb source itself in that case ?

i don't quite see what this changes—folks who don't have bash installed will still see failures; it's only the nature of the failure that will be different (`run` step will fail if `shell:` is set to a shell that isn't installed, vs. a manual `exit code 1` that actually tells users what happened)...

iynere:
i guess i don't see how this helps the orb work w/ alpine images @zzak installing bash in your config fiile should also enable the orb to work w/ those images as-is.... or were you finding that even after installing bash, the `run` steps would still default to `sh` when the orb is used w/ alpine-based images ?

iynere:
ahhh, i see 😄 this seems like a good fix then ! @zzak

iynere:
how's it now ? @zzak 

i split the difference

zzak:
@iynere LGTM, thank you for your help on this!

PR #24, opened 2019-01-02
improve formatting for `curl` commands

gsaslis:
at the moment, the curl commands that invoke the slack API are formatted as a single-line command, making it harder to read / follow, especially with all the escaped quotes, etc. 

I think splitting this into multi-line, properly-indented json makes it much easier to follow / maintain / spot any issues.


iynere:
thank you ! @gsaslis

Issue #16, opened 2018-12-20
shell verification shouldn't be based on $SHELL


CaptainPatate:
Depending on the Docker image we use, this orb can fail even if `bash` is installed.
For example with a Docker image like this:
```Dockerfile
FROM alpine:3.7

RUN apk --no-cache add bash\
                       ca-certificates\
                       make

WORKDIR /app

ENTRYPOINT ["make"]
```

```shell
docker run --rm -it --entrypoint /bin/sh captainpatate/make-dnscontrol:v0.2.7 
/app # echo $SHELL

/app # echo $BASH

```
```shell
docker run --rm -it --entrypoint /bin/bash captainpatate/make-dnscontrol:v0.2.7 
bash-4.4# echo $SHELL
/bin/ash
bash-4.4# echo $BASH
/bin/bash
```

Even if we use `bash`, I guess some Docker entrypoint trick makes `$SHELL` report the login shell instead of the real shell. It would be better to use the presence of the variable `$BASH` to verify that we are indeed inside a Bash shell.

zzak:
Hello @CaptainPatate,

Thanks for your report.

It seems that bash reserves the environment variable `$BASH` for this reason (inside bash):
http://tldp.org/LDP/Bash-Beginners-Guide/html/sect_03_02.html#sect_03_02_04_02

I'm wondering though if we can rely on it?

Would you be willing to send a patch to replace the shell check?

CaptainPatate:
Yes sure, I'll try it next week and make a PR if it works.

PR #27, opened 2019-01-07
Adding job for notifying about awaiting approval

tommybo-entur:
using image spotify/alpine (can change this to an alpine image from circleci if needed and available). This is also changing the shell-check from using $SHELL to $BASH as suggested in issue #16.

### DIFF
--- a/src/@orb.yml
+++ b/src/@orb.yml
@@ -17,7 +17,7 @@ description: |
 executors:
   alpine:
     docker:
-      - image: spotify/alpine:latest
+      - image: cibuilds/base:latest
         environment:
           TERM: dumb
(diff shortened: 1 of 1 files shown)


KyleTryon:
I like the idea of including an approval job. @iynere actually has a similar system already that is similar to this. I have asked her to take a look and see if we can add that on top of that you have here.

For the executors, I believe we actually do not have a CircleCI base alpine image. I will investigate why that is again. For now, I don't see any problem with using Spotify's image, that looks very useful.

I plan to no longer make bash a requirement shortly, but I'd like to get this error handling in the meantime now. Thank you!

PR #30, opened 2019-02-01
Add parameter which only for tag to status command

kitakitabauer:
### Checklist

### Description
Added a new parameter - only_for_tag - to the slack/status command, which allows slack updates to be triggered for builds on any tag only.

### DIFF
--- a/src/commands/status.yml
+++ b/src/commands/status.yml
@@ -54,9 +54,9 @@ steps:
       shell: /bin/bash
       when: always
       command: |
-        if [[ "x" == "x<< parameters.only_for_branch >>" && "${CIRCLE_BRANCH}" == "<< parameters.only_for_branch >>" ]] ||
-           << parameters.only_for_tag >> && [ -n "${CIRCLE_TAG}" ] ||
-           [[ -z "<< parameters.only_for_branch >>" && ! << parameters.only_for_tag >> ]]; then
+        if [[ -n "${CIRCLE_BRANCH}" && "x" == "x<< parameters.only_for_branch >>" && "${CIRCLE_BRANCH}" == "<< parameters.only_for_branch >>" ]] ||
+           [[ -n "${CIRCLE_TAG}" && "xtrue" == "x<< parameters.only_for_tag >>" ]] ||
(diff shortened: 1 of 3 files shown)


src/commands/status.yml
```diff
@@ -49,7 +54,9 @@ steps:
       shell: /bin/bash
       when: always
       command: |
-        if [ "x" == "x<< parameters.only_for_branch>>" ] || [ "${CIRCLE_BRANCH}" == "<< parameters.only_for_branch>>" ]; then
+        if [[ "x" == "x<< parameters.only_for_branch >>" && "${CIRCLE_BRANCH}" == "<< parameters.only_for_branch >>" ]] ||
+           << parameters.only_for_tag >> && [ -n "${CIRCLE_TAG}" ] ||
```
eddiewebb: I didn't think bash will evaluate this as true/false, and needs to check for matching the literal string "true" or false is empty..

iynere:
@kitakitabauer thanks, this is a good feature add. i have some thoughts about doing this in a more agnostic way, so let me take a stab at this. i may push my own branch & PR—if so, i'll CC you on it so you can 👀

kitakitabauer:
@iynere thanks, i'm waiting for your PR!

src/commands/status.yml
kitakitabauer: thanks! condition changed. d00f869

kitakitabauer:
@iynere When will you be able to PR? This function is necessary as soon as possible.

iynere:
@kitakitabauer sorry, was working on a higher-level solution for this that would be applicable to any orb, but it got complicated ! i'll prioritize merging a specific solution for this issue over my broader implementation, thanks for the ping !

iynere:
@kitakitabauer i ended up implementing my solution—can you refactor this PR to use it?

i've added a command to circleci's build-tools-orb that allows you to run any arbitrary command or set of commands only on certain branches or tags:

https://circleci.com/orbs/registry/orb/circleci/build-tools#commands-run-filtered

with that, you should be able to do what you're trying to do here

KyleTryon:
+1 love the solution implemented here. Should be highly extendable.

KyleTryon:
Outdated.

PR #31, opened 2019-02-10
Add more parameters to further customize messages from the notify command

rmacklin:
### Checklist

  - I didn't think that adding these optional parameters to the existing notify command was considered a significant new feature, so I didn't add new examples. It seems like the README additions should be sufficient, but let me know if you disagree.

### Motivation

For our needs, we wanted to send some messages without the _Project_ and _Job Number_ fields and the _Visit Job_ action. We also wanted to send some messages with an author, title, and footer in the message attachment. This PR adds support for doing that in a backwards-compatible manner. I've published these changes under `rmacklin/slack@1.1.0` and verified that without specifying any of the new parameters the behavior is the same as before.

### Description

- Add additional optional parameters to the `slack/notify` command

  Specifically, add support for specifying the `author_name`, `author_link`, `title`, `title_link`, `footer`, and `ts` properties of a Slack message attachment (https://api.slack.com/docs/message-attachments)

  Each of these new parameters defaults to the empty string, which makes them ignored by the Slack API unless they're explicitly overridden to a nonempty value.

- Add boolean parameters to control inclusion of the `fields` and `actions` of the `slack/notify` command

  By default, messages from the `slack/notify` command have the _Project_ and _Job Number_ fields and the _Visit Job_ action. These can now be removed by setting `include_project_field`, `include_job_number_field`, and/or `include_visit_job_action` to `false` in the command parameters.


rmacklin:
@iynere @KyleTryon Do you have any thoughts on this PR?

iynere:
this looks awesome, thank you ! @rmacklin

iynere:
its bugging me that this readme doesn't fully use markdown tables for params, so im must gonna push a few commits & then merge @rmacklin

rmacklin:
Sounds good, thanks!

iynere:
@KyleTryon can you give this readme a quick 👀 & let me know if everything's 👌 ? 🙏

KyleTryon:
I am extremely happy with this. The docs look much better this way too. I wonder if maybe (now or in the future) we should add a column for optional parameters so new users can easily determine the minimum needed to get started. If slack ignores the empty string that makes it much easier on us, should cause no issues. I'm fine with merging @iynere

README.md
```diff
@@ -57,12 +69,15 @@ jobs:
...
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `webhook` | `string` | ${SLACK_WEBHOOK} | Enter either your webhook value or use the CircleCI UI to add your token under the `SLACK_WEBHOOK` environment variable |
+| `mentions` | `string` |  | Comma-separated list of Slack User or Group (SubTeam) IDs (e.g., "USER1,USER2,USER3"). _**Note:** these are Slack User IDs, not usernames. The user ID can be found on the user's profile. Look below for information on obtaining Group ID._ |
+| `fail_only` | `string` | false | If set to "true," successful jobs will _not_ send alerts |
+| `only_for_branch` |  | ${SLACK_WEBHOOK} | If set, a specific branch for which slack status updates will be sent |
```
rmacklin: Since there is no default for `only_for_branch:
https://github.com/CircleCI-Public/slack-orb/blob/3057745546bd905ff4eaca2c7682452ebab823df/src/commands/status.yml#L21-L24

should this be:
```suggestion
| `only_for_branch` |  |  | If set, a specific branch for which slack status updates will be sent |
```
?

iynere: ugh, oops, thank you !

rmacklin:
Thank you @iynere and @KyleTryon! From https://circleci.com/orbs/registry/orb/circleci/slack it looks like this is now published under `circleci/slack@2.0.0`. Are there any breaking changes I should be aware of if I'm upgrading from version 1.0?

iynere:
not that i know of ! @rmacklin but try it & let us know ?

rmacklin:
Ok, I will. Just wasn't sure if there was a specific reason it was published with a new major version rather than a new minor version.

iynere:
@rmacklin the version bumps are dynamically interpolated based on `git diff` analysis : https://github.com/CircleCI-Public/orb-tools-orb/blob/master/src/jobs/trigger-integration-workflow.yml#L106

so, any modifications to jobs / commands trigger a major version bump

arguably, that's a bit overzealous, but that's how it's setup for now

feel free to PR that file in that repo if you have a better idea, though ! @rmacklin 

any changes there will have an extremely powerful ripple effect on all our orbs, since almost all of them use that dynamic `git diff` interpolation to decide what kinds of releases to publish

my thoughts are, we should check for NEW files, rather than just any modifications, in order to publish a major release, & keep modifications for minor / patch releases

rmacklin:
Ah, I see. I didn't know the version bump was managed automatically like that. Thanks for explaining!

iynere:
this issue came up in another situation today, as well, so I made a PR to address it @rmacklin 

https://github.com/CircleCI-Public/orb-tools-orb/pull/20

PR #32, opened 2019-03-01
Update orb version in readme

owensbla:


iynere:
the examples aren't meant to reflect the latest version @owensbla 

`1.0.0` is just a simple, easy-to-undrestand string. thank you tho !

owensbla:
Might want to make it a little more clear that the certain parameters in the commands are only available in 2.0.0 then.

I updated it because I only saw 1.0.0 referenced everywhere then tried using notify with parameters like title, title_link, include_project_field, etc. and had to debug it to figure out it was just a version thing and those were part of 2.0.

iynere:
hmmmm..... maybe we should just use x.y.z in the examples then, what do you think ? @felicianotech

felicianotech:
Shouldn't the commands and such not be in the readme at all since the registry shows them live? Listing everything in the readme could get out of date quickly.

iynere:
that doesn't address the original question @felicianotech 

i generally agree, but seeing that info in the repo is nice too. but yes, it's very difficult to keep up to date.

iynere:
perhaps we could write an orb to automate creating the markdown for the readme ?

Issue #33, opened 2019-03-08
slack/notify on fail?


ahtutejlo:
How to run slack/notify on fail builds only?
I can run it in slack/status but it is not possible to set a custom message there.

cayter:
Any plan to add `when: <on_fail|on_success>` support?

KyleTryon:
Hello @cayter 

Would the status command not work for you? https://github.com/CircleCI-Public/slack-orb#status

cayter:
@KyleTryon Nope as we needed to support custom message.

lokst:
Closing this issue as it looks like it is already resolved by https://github.com/CircleCI-Public/slack-orb/pull/37

PR #34, opened 2019-03-13
Add `when` attribute for `slack/notify` command

cayter:
### Checklist

### Motivation, issues

Related issue: https://github.com/CircleCI-Public/slack-orb/issues/33

### Description

This PR is to introduce `when` attribute to `slack/notify` so that users can freely decide when to send the notification based on CircleCI's job status.

### DIFF
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -75,13 +75,13 @@ steps:
   - run:
       name: Slack - Setting Failure Condition
       command: |
-        echo 'export SLACK_BUILD_STATUS="on_fail"' >> $BASH_ENV
+        echo 'export CIRCLE_BUILD_STATUS="on_fail"' >> $BASH_ENV
       when: on_fail
 
   - run:
       name: Slack - Setting Success Condition
       command: |
-        echo 'export SLACK_BUILD_STATUS="on_success"' >> $BASH_ENV
+        echo 'export CIRCLE_BUILD_STATUS="on_success"' >> $BASH_ENV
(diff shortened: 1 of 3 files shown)


iynere:
this is a great idea ! @cayter 

can i suggest you change the parameter type from `string` to `enum` ?

https://circleci.com/docs/2.0/reusing-config/#enum

that way, we only give folks options that are actually available

thank you !

rdublenych:
Any news on this?

KyleTryon:
Hello @rdublenych Thank you for contributing. Would you mind pushing another commit in order to trigger our automated testing? I believe the branch may be out of date so we are missing the statuses at the moment.

cayter:
@rdublenych @KyleTryon @iynere Can help with reviewing the latest change? Just got time to get back to this. Sorry and thanks.

src/commands/notify.yml
```diff
@@ -85,6 +103,12 @@ steps:
           echo "Please input your SLACK_WEBHOOK value either in the settings for this project, or as a parameter for this orb."
           exit 1
         else
+          if [[ -n "<< parameters.when >>" && << parameters.when >> != "always" ]]; then
```
KyleTryon: I believe this step will need to be modified so this step itself possesses the "when:always" parameter. If the job has failed before the notification this step will not execute, only the setting failure condition step would have the chance to run.

KyleTryon: @cayter ^

PR #36, opened 2019-03-21
Copy include_* options from notify to status

stig:
### Checklist

- [X] All new jobs, commands, executors, parameters have descriptions
- [X] Examples have been added for any significant new features
- [X] README has been updated, if necessary

### Motivation, issues

We do a lot of deployments across a lot of environments and while the status updates are valuable, they are currently very verbose. We would like them to fit on a single line.

### Description

Copied include_* from notify to status command


stig:
I've rebased this against current master. Would love feedback on whether you're likely to accept this.

iynere:
@stig is your fork building on circleci ? if not, can you add it, and then temporarily modify the deployment target for the `publish-dev` job in the config file on your fork, so it publishes a dev version to a namespace/orb you have publish rights to ?

i want to see an example of this modification in action. thanks so much & apologies for the extra hoops to jump through ! i'm following up internally to see if we can figure out an easier way to enable folks' forked PRs to publish dev versions to one of our namespaces, without passing our internal API tokens down to those jobs....

iynere:
ok, just kidding, worked out a better solution for now @stig 

i've now set up a multi-branch system here—i'll merge this PR to our staging branch, which will publish a dev orb you can try out in your jobs to ensure it does what you want it to

please do that & drop a link to a job where you've used it, & maybe a screenshot of the message in slack, & assuming all the integration tests pass on `staging`, we can get this merged

thanks for your patience !

PR #40, opened 2019-04-10
Support here,channel,everyone mentions

fumikony:
### Checklist

- [x] All new jobs, commands, executors, parameters have descriptions
- [x] Examples have been added for any significant new features
- [x] README has been updated, if necessary

### Motivation, issues

We want to use `@here` in mentions.

### Description
Add support for channel mentions: `@here`, `@channel` and `@everyone`.

PR #43, opened 2019-04-16
fix mentions

fumikony:
### Checklist

### Motivation, issues
pull request #40 breaks normal mentions if slack user ids include `here`, `channel` or `everyone`.

### Description
- fix the pattern for `grep -E`.
- add a test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,12 @@ jobs:
       - slack/approval:
           mentions: "here"
 
+  approvaltest-mention-here-2:
+    executor: ci-base
+    steps:
+      - slack/approval:
+          mentions: "foo-here-bar"
+
 # yaml anchor filters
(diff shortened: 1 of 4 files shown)


iynere:
@fumikony what is the expected behavior w/ this PR ? the test you added came out like this, which doesn't seem right to me, unless i'm misunderstanding...

image

fumikony:
@iynere thanks for reviewing.
The test result is right.

Before merging this PR, a result of `approvaltest-mention-here-2` is like this;
スクリーンショット 2019-04-17 17 04 26

This shows that `foo-here-bar` is mistakenly treated as `here`.

After merging, the result is like this;
スクリーンショット 2019-04-17 17 21 47

This shows `foo-here-bar` is treated as a normal user id.

iynere:
@fumikony ahhh, gotcha, thank you !!! sounds good :)

PR #49, opened 2019-04-19
commands/status: Set success_message / failure_message to fallback field

### LINKED ISSUE #47: Using `failure_message` in `status` seems to set only the `text` field of attachment, not the `fallback` one
Using `failure_message` in `status` seems to set only the `text` field of attachment, not the `fallback` one.

It is a problem when using it with bots such as `hubot-slack`.

### Orb version

2.2.0

### What happened

I'm using the `slack/status` command with `failure_message`.
It seems that the current implementation sets only the `text` field of attachment, not the `fallback` one.

https://github.com/CircleCI-Public/slack-orb/blob/563d26ac70dc5989704455b4db12e8ab27599517/src/commands/status.yml#L147

It is a problem when using it with bots such as `hubot-slack`.

The `hubot-slack` does not respond message sent by the `status` command because it parses the `fallback` field if a message has an attachment.

https://github.com/slackapi/hubot-slack/blob/e3acc28f224cb8fb60b2288b1395c25c063b27b2/src/message.coffee#L104-L109

It's a very confusing behavior.
I wasted time thinking that the regular expression parsing the message was wrong.

### Expected behavior
The `text` and `fallback` field should set to the same value or contain the `failure_message` parameter.

It may be a `hubot-slack` problem, but the `notify` command can set the `message` in the `fallback` field, but it does not seem to equal the `text` one.

https://github.com/CircleCI-Public/slack-orb/blob/563d26ac70dc5989704455b4db12e8ab27599517/src/commands/notify.yml#L108-L109

How to fix it?
I will be happy to fix it if we can agree on the policy.

Thanks!

minamijoyo:
### Checklist

### Motivation, issues

Fixes #47

In the current implementation, using `failure_message` in status doesn't
affect the `fallback` field.
It causes a problem when sending a message to bots such as `hubot-slack`.
The `hubot-slack` parses the `fallback` field if a message has an attachment.

### Description

We should set `text` and `fallback` field to the same value.
The `success_message` also have the same issue. So I fixed both.

Any feedback is welcome!

### DIFF
--- a/src/commands/status.yml
+++ b/src/commands/status.yml
@@ -105,7 +105,7 @@ steps:
                   --data "{ \
                             \"attachments\": [ \
                               { \
-                                \"fallback\": \"A job has succeeded - $CIRCLE_BUILD_URL\", \
+                                \"fallback\": \"<< parameters.success_message >>\", \
                                 \"text\": \"<< parameters.success_message >>\", \
                                 \"fields\": [ \
                                   <<# parameters.include_project_field >>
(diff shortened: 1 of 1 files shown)


minamijoyo:
Thanks 😄

PR #58, opened 2019-06-01
Wrap `status` command with sensible defaults

zzak:
We use this command a lot, and everywhere we essentially repeat the
following over and over. It would be great if the orb just provided it
for us!

```
orbs:
  slack: circleci/slack@2.5.1

commands:
  # Wrap the slack/status command to only notify on failures
  slack-notify-on-failure:
    steps:
      - slack/status:
          fail_only: "true"
          only_for_branch: "master"
```

### Checklist

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,6 +1,0 @@
-extends: relaxed
-
-rules:
-    line-length:
-        max: 200
-        allow-non-breakable-inline-mappings: true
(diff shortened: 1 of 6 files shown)


zzak:
Opening this for a sanity check, @iynere what do ya'll think?

iynere:
looks nice, should keep things DRY @zzak thank you

for things like this, we need to manually publish a `dev:alpha` version first from someon'es local machine in order for the CI workflow to proceed, i usually include a shell script file to do it in every orb repo but looks like i never added one here, i'll add that to your branch

zzak:
Thanks @iynere!!

PR #80, opened 2019-09-30
upload.yml added usinf curl

sayanmondal2098:
### Checklist


KyleTryon:
This is not required as the step itself will/should check for bash.

PR #81, opened 2019-10-01
Fix check curl exists

### LINKED ISSUE #25: Step reports success even if alert fails.
screen shot 2019-01-03 at 6 49 40 pm

My relevant config:

```
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,8 @@
 version: 2.1
+
+orbs:
+  slack: circleci/slack@0.1.15
+
 jobs:
   build:
     docker:
@@ -23,6 +27,8 @@ jobs:
+      - slack/status:
+          fail_only: 'true'
```

piyush-insider:
### Checklist

### Motivation, issues
Fixes #25 

### Description
Using the which utility checking if cURL exists, if it does the code returned is 0 else the code returned is 1 by the command, Which we then check with grep and in case of 1 we exit and echo `curl not installed`

### DIFF
--- a/src/commands/approval.yml
+++ b/src/commands/approval.yml
@@ -44,6 +44,9 @@ steps:
           echo Bash not installed.
           exit 1
         fi
+        # Provide error if curl is not installed
+        which curl > curl_exists; echo $? | grep -q '1' && echo curl not installed && rm out && exit 1
+        rm curl_exists
         # Provide error if no webhook is set and error. Otherwise continue
         if [ -z "<< parameters.webhook >>" ]; then
           echo "NO SLACK WEBHOOK SET"
--- a/src/commands/approval.yml
+++ b/src/commands/approval.yml
@@ -45,11 +45,6 @@ steps:
       name: Slack - Sending Approval Notification
       shell: /bin/bash
       command: |
-        # Provide error if non-bash shell
-        if [ -z "$BASH" ]; then
-          echo Bash not installed.
-          exit 1
-        fi
         # Provide error if curl is not installed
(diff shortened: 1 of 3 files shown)


piyush-insider:
@KyleTryon Thanks for the review. I have added curl as a step in all the three commands. As for the suggestion of having all prerequisite checks in a single step, I think that should be a part of a separate refactor and should not go in along with this change. I'm filing an issue for it separately and will try to pick it up as soon as possible. What do you think?

src/commands/approval.yml
```diff
@@ -41,6 +41,12 @@ steps:
           exit 1
         fi
 
+  - run:
+      name: Provide error if cURL is not installed.
+      command: |
+        which curl > curl_exists; echo $? | grep -q '1' && echo curl not installed && rm curl_exists && exit 1
```
sachincool: @piyush-insider try implementing this it looks better and without any External file creation. https://scripter.co/check-if-a-command-exists-from-shell-script/

piyush-insider: Sure. Will take a look at it

KyleTryon:
I have closed issue #25 to reduce the duplicate PR submissions. This PR will be our active branch for this fix.

src/commands/approval.yml
KyleTryon: It looks as if once this change is added we should be ready to merge.

wyardley: `command -v` should be a good way to do it too, I think.

eddiewebb:
@KyleTryon - I suggest we merge `[skip-ci]` to staging, and we can mass promote all staging additions end of month?

Issue #78, opened 2019-09-27
[Request] Support for message-attachments "image_url"


strangeminds:
We would like to add images to our CircleCI slack notifications using the orb.  Right now when we include a giphy URL to the message, it does not render the gif in Slack.

Request to add support for Slack message-attachment **image_url**

https://api.slack.com/docs/message-attachments

```
version: 2.1
orbs:
  slack: circleci/slack@3.3.0
jobs:
  fireworks:
    machine: true
    steps:
      - slack/notify:
          message: "https://giphy.com/gifs/fireworks-IjmMzurYulKEw"
          webhook: ${SLACK_WEBHOOK}
```

steevsachs:
I'm digging into CircleCI and this seems like a good first issue.  Mind if I take a crack at this?

PR #90, opened 2019-10-02
Add support for image_url on notifications

steevsachs:
Adds support for `image_url` passed to `notify` in order to attach an image to the notification message.

Reference CircleCI-Public/slack-orb#78


KyleTryon:
@steevsachs Sorry for the delay. Thanks for the awesome addition!

PR #95, opened 2019-10-03
Checkcurl

sayanmondal2098:
### Checklist


KyleTryon:
Hello thank you for contributing but a bash check in the example documentation should not be required. This is automatically checked for you via the commands themselves and does not need to be added by the user.

Issue #100, opened 2019-10-08
[Request] Support all parameters for Runtime customizations


yukihira1992:
I want to  support all parameters in addition to the `channel` parameter of incoming webhook. #67

- `username`
- `icon_emoji`
- `icon_url`

https://api.slack.com/custom-integrations/incoming-webhooks

KyleTryon:
I just discovered this note on slack's docs that may warrant further research:

https://api.slack.com/incoming-webhooks

> You cannot override the default channel (chosen by the user who installed your app), username, or icon when you're using Incoming Webhooks to post messages. Instead, these values will always inherit from the associated Slack app configuration.

PR #67, opened 2019-07-24
Add `channel` parameter for overring webhook's default

rail44:
### Checklist

- [x] All new jobs, commands, executors, parameters have descriptions
- [x] Examples have been added for any significant new features
- [x] README has been updated, if necessary

### Motivation, issues

refs #53 

Slack webhook has `channel` parameter passed by POST data.
It overrides slack channel configured when webhook created.
With this, we can make notification for various channel with single webhook url.
<!---
	why is this change required? what problem does it solve?
  paste links to any relevant GitHub issues filed against
  this repository that this pull request addresses
--

PR #101, opened 2019-10-08
Add parameters for incoming webhook's runtime customizations. #100

yukihira1992:
### Checklist

### Motivation, issues

refs #100 

In addition to `channel` parameter supported in #67, `username`, `icon_emoji` and `icon_url` parameters are available.
https://api.slack.com/custom-integrations/incoming-webhooks


KyleTryon:
I just discovered this note on slack's docs that may warrant further research:

https://api.slack.com/incoming-webhooks

> You cannot override the default channel (chosen by the user who installed your app), username, or icon when you're using Incoming Webhooks to post messages. Instead, these values will always inherit from the associated Slack app configuration.

PR #104, opened 2019-10-08
feat: Allow suppressing certain fields for approvals

wyardley:
### Checklist

### Motivation, issues
This addresses an inconsistency, where the project and job number fields can't be suppressed when sending an approval message, as they can in other contexts.

Let me know if there's a way to auto-generate the `README` (there should be if there isn't!), or if you want me to squash to one commit. Since the docs were missing entirely, I made two separate commits.

### Description
* Support suppressing `include_project_field` and `include_job_number_field` for the `approval` command and `approval-notification` job, similar to what's supported for other commands
* Add missing docs for the above types, including information about the new parameters

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,36 @@ jobs:
 
 ## Commands
 
+### Approval
+Send a notification that a manual approval job is ready
+
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `color` | `string` | '#3AA3E3' | Hex color value for notification attachment color. |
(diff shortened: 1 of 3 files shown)


wyardley:
I assume the publish failure is due to me being an outside contributor, but would be nice if the workflow could be broken up a little differently?

KyleTryon:
Hello @wyardley We are currently in the midst of planning a new pipeline process to avoid this kind of issue. Along with likely a standardized testing method for orbs.

This does look good to me though and we'll merge this in 👍  Thank you

PR #109, opened 2019-10-11
image Notification

ritwickbhargav80:
### Checklist

- [x] Image as notification on Slack 

### Motivation, issues
It's needed because it was not defined in the earlier version.

### Description
Missing image as a parameter and not defined


KyleTryon:
Not a valid parameter for the job or commands.

PR #110, opened 2019-10-11
add `change_only` flag to status command

### LINKED ISSUE #54: Support for changed build statuses
### Feature request

We do many builds a day, and use Slack for visibility on the status of those builds. A message every time the build succeeds is not particularly useful information, as this is the default, and the significant majority of our builds succeed. Generally, we want notifications for:

- Build failures
- A build success the first time after a failure

The former tells us we need to go and fix something, and continued notifications of failures make it less likely that it will be forgotten. The latter notifies those who may be interested in the project, but not involved in the fixing of the build, that everything is ok once again.

It would be great to have the notion of whether the build status has _changed_ from the last run, to enable workflows like this.

imnotjames:
### Checklist

### Motivation, issues

Closes #54 

In some cases, users will want to only be notified of status **changes** instead of all Ci statuses.

### Description

Add `change_only` flag to command `status`.

When disabled, it has no impact.

When enabled, the status is reported to slack only on `fixed` or `broke` statuses.
For example, with the following statuses over time:

Status | Experience
------|------------
Pass | No Alert
Pass | No Alert
Pass | No Alert
Fail | Send Alert
Fail | No Alert
Fail | No Alert
Fail | No Alert
Pass | Send Alert
Fail | Send Alert
Fail | No Alert
Pass | Send Alert
Pass | No Alert
Pass | No Alert
Pass | No Alert

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Send a status alert at the end of a job based on success or failure. This must b
 | `include_job_number_field` | `boolean` | `true` | Whether or not to include the _Job Number_ field in the message |
 | `include_visit_job_action` | `boolean` | `true` | Whether or not to include the _Visit Job_ action in the message |
 | `channel` | `string` | | ID of channel if set, overrides webhook's default channel setting |
+| `circle_token_name` | `string` | `CIRCLE_TOKEN` | The name of environment variable containing CircleCI API Token. Only required for private projects when `change_only` is set to true. |
 
 Example:
 
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ Send a status alert at the end of a job based on success or failure. This must b
 | `failure_message` | `string` | :red_circle: A $CIRCLE_JOB job has failed! $SLACK_MENTIONS | Enter your custom message to send to your Slack channel |
 | `mentions` | `string` |  | Comma-separated list of Slack User or Group (SubTeam) IDs (e.g., "USER1,USER2,USER3"). _**Note:** these are Slack User IDs, not usernames. The user ID can be found on the user's profile. Look below for information on obtaining Group ID._ |
 | `fail_only` | `boolean` | `false` | If set to `true`, successful jobs will _not_ send alerts |
+| `change_only` | `boolean` | `false` | If set to `true`, consecutive successes or failures will _not_ send alerts |
 | `only_for_branches` | `string` |  | If set, a comma-separated list of branches, for which to send notifications |
 | `include_project_field` | `boolean` | `true` | Whether or not to include the _Project_ field in the message |
 | `include_job_number_field` | `boolean` | `true` | Whether or not to include the _Job Number_ field in the message |
@@ -136,6 +137,7 @@ jobs:
       - slack/status:
           mentions: "USERID1,USERID2" # Optional: Enter the Slack IDs of any user or group (sub_team) to be mentioned
           fail_only: true # Optional: if set to `true` then only failure messages will occur.
+          change_only: true # Optional: if set to `true` then only "change" in status messages will occur.
           webhook: "webhook" # Optional: Enter a specific webhook here or the default will use $SLACK_WEBHOOK
           only_for_branches: "master" # Optional: If set, a specific branch for which status updates will be sent. In this case, only for pushes to master branch.
 ```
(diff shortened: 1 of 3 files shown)


imnotjames:
Ok!  This might be good now.  I've tested it locally, and deployed it as an orb at `imnotjames/slack` but I have no idea how to link to that here. :)

imnotjames:
I'd like to mention that I'm not a fan of adding more nested `if` blocks but I chose to do that to maintain consistency with existing code style over what I would consider as more readable and cleaner.

imnotjames:
Eep!  You're very right! I misunderstood something else I had been reading, and in testing I was running under the assumption that those existed.

I'll take a look at the API and see what I can do to make this happen.

imnotjames:
This should pull from the API now at this point - but I've yet to test this LIVE on a circle CI build.  I've tested this locally using `circleci local execute`

README.md
```diff
@@ -111,11 +111,13 @@ Send a status alert at the end of a job based on success or failure. This must b
...
 | `fail_only` | `boolean` | `false` | If set to `true`, successful jobs will _not_ send alerts |
+| `change_only` | `boolean` | `false` | If set to `true`, consecutive successes or failures will _not_ send alerts |
 | `only_for_branches` | `string` |  | If set, a comma-separated list of branches, for which to send notifications |
 | `include_project_field` | `boolean` | `true` | Whether or not to include the _Project_ field in the message |
 | `include_job_number_field` | `boolean` | `true` | Whether or not to include the _Job Number_ field in the message |
 | `include_visit_job_action` | `boolean` | `true` | Whether or not to include the _Visit Job_ action in the message |
 | `channel` | `string` | | ID of channel if set, overrides webhook's default channel setting |
+| `circle_token_name` | `string` | `CIRCLE_TOKEN` | The name of environment variable containing CircleCI API Token. Only required for private projects when `change_only` is set to true. |
```
KyleTryon: Here we'd want to use the "env_var_name" type: https://circleci.com/docs/2.0/reusing-config/#environment-variable-name

To hopefully prevent users from using their key in plain text.

src/commands/status.yml
```diff
@@ -26,6 +26,12 @@ parameters:
     description: >
       If `true`, notifications successful jobs will not be sent
 
+  change_only:
+    type: boolean
+    default: false
+    description: >
+      If `true`, notifications that are not "broke" or "fixed" will not be sent.
```
KyleTryon: can we change 'broke' to 'failed'?

src/commands/status.yml
```diff
@@ -63,6 +69,12 @@ parameters:
     description: >
       ID of channel if set, overrides webhook's default channel setting
 
+  circle_token_name:
+    default: CIRCLE_TOKEN
+    description: The name of environment variable containing CircleCI API Token.
+      Only required for private projects when `change_only` is set to true.
+    type: string
```
KyleTryon: env_var_name parameter type

src/commands/status.yml
```diff
@@ -167,46 +207,51 @@ steps:
...
-                          \"url\": \"$CIRCLE_BUILD_URL\" \
-                        } \
-                        <</ parameters.include_visit_job_action >>
-                      ], \
-                      \"color\": \"#ed5c5c\" \
-                    } \
-                  ] \
-                } " << parameters.webhook >>
-              echo "Job failed. Alert sent."
+              if [ << parameters.change_only >> = true ] && [ "success" != "$SLACK_LAST_BUILD_STATUS" ]; then
```
KyleTryon: Has the outcome been tested for jobs that have been skipped or canceled?

imnotjames: Looking into this a bit..

The current slack orb counts `success` as a success, and anything else as a failure.

This means that the following are counted as a failure:

* `canceled`
* `infrastructure_fail`
* `timedout`
* `failed`
* `no_tests`

If we leave with the same logic as it stands it would be the current behavior in the code.  It doesn't seem like the the current orb differentiates between the different failure scenarios.

We _could_ check if it's EXACTLY a failure and in that case it would emit erroneously at some points but that might be counted as acceptable (eg, I'm not sure if `when: on_fail` is fired for something like `timedout`)

src/commands/status.yml
imnotjames: Sure, I think I used the terminology from another orb / iteration.

src/commands/status.yml
KyleTryon: We are going to close this for now and revisit the topic when we have a more deterministic method of determining changes. We will continue to look into this not only for Slack but this same mechanic is useful in mono-repos.

KyleTryon:
We are going to close this for now and revisit the topic when we have a more deterministic method of determining changes. We will continue to look into this not only for Slack but this same mechanic is useful in mono-repos.

Issue #128, opened 2019-11-04
Add Then Replace Executor


felicianotech:
This orb includes a single executor called `alpine`. It uses the Docker image`cibuilds/base`, which is one of my Alpine-based images from my `cibuilds` project. We now have a suitable replacement, `cimg/base`.

According to our [Orbs Best Practices](https://circleci.com/docs/2.0/orbs-best-practices/#executors), we should have a default executor names `executor`, which this orb currently does not.

The route I see forward here is:

- [ ] PR #129 - add a `default` executor that contains the `cimg/base` image
- [ ] PR #130 - remove the `alpine` executor in the future

KyleTryon:
Note: Ensure the `stable` tag is used.

dsayling:
To be fixed in v4. Leaving open until addressed.

PR #129, opened 2019-11-04
Add cimg/base as default executor.

felicianotech:
Related #128


KyleTryon:
Not sure how the first two checks are missing but it clearly ran 👍 
https://github.com/CircleCI-Public/slack-orb/pull/129/checks?check_run_id=288285065

PR #130, opened 2019-11-04
Remove old executor.

felicianotech:
Related #128.

This PR should likely be merged sometime after #129 to allow people a transition time.


KyleTryon:
Going to save this one a bit longer til the pipeline is updated so we can ensure the version increment wont accidentally bump users who are still on Alpine. This repo is still using the non [semver:segment] pipeline.

gmemstr:
While I think this should be done sooner, we'll go ahead and wrap this up in #160

Issue #124, opened 2019-10-29
Unable to send from tag build


jodastephen:
### Orb version

v3.4.0

### What happened

Error "Current branch is not included in only_for_branches filter; no status alert will be sent" is output. This occurs with config:

```
      - slack/status:
          channel: XXXXXXX
          failure_message: 'failed'
          fail_only: true
          only_for_branches: master
```

This happens because on a tag build, `$CIRCLE_BRANCH` is not set, so the ORB fails to correctly  match the branch name.

### Expected behavior

The slack message should be sent.

BeyondEvil:
@jodastephen I'm using this workaround:

`only_for_branches: ",develop,master"`

gmemstr:
@KyleTryon Does this apply to version 4.x?

PR #132, opened 2019-11-27
Support sending notifications on tags (#124)

BeyondEvil:
Fixes: #124

### Checklist

### Motivation, issues

To send notifications on a tagged build, you would have to use a hack (see my comment in #124).

Apart from not having to rely on hack, this gives the user a more flexible solution instead of just an on/off-switch.

### Description

Add `use_on_tag` to `status` (and `notify-on-failure`) to enable users to push slack messages on tags with a greater level of granularity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ Send a status alert at the end of a job based on success or failure. This must b
 | `mentions` | `string` |  | Comma-separated list of Slack User or Group (SubTeam) IDs (e.g., "USER1,USER2,USER3"). _**Note:** these are Slack User IDs, not usernames. The user ID can be found on the user's profile. Look below for information on obtaining Group ID._ |
 | `fail_only` | `boolean` | `false` | If set to `true`, successful jobs will _not_ send alerts |
 | `only_for_branches` | `string` |  | If set, a comma-separated list of branches, for which to send notifications |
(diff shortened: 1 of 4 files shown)


BeyondEvil:
Couldn't quite figure out the tests. Will need some hand-holding if you require me to add any. 🙏

BeyondEvil:
> So the purpose of this addition seems to be to allow a user to automatically send a message on a tagged build. However, as the user will already have a workflow with a tag filter on it, would it not make more sense to give separate parameters to that job with the existing orb notify and status commands?

I don't know, maybe?

In our case, we only have one workflow. We use `master` branch for stage-releases and tags for production-releases. The flows are the same, just using different environments.

But maybe we're using Circle CI incorrectly? 🤷‍♂

BeyondEvil:
ping @KyleTryon

KyleTryon:
Hello @BeyondEvil,

Sorry for the lack of communication. We have been hard at work on the 4th version of the Slack orb, now available here: https://circleci.com/developer/orbs/orb/circleci/slack

Here is a usage example using the new orb showing how you might send a notification on a tagged commit: https://circleci.com/developer/orbs/orb/circleci/slack#usage-successful_tagged_deployment

If this does not satisfy your use-case please let me know in a new issue and we would be happy to look into solutions for the new orb. Thank you!

Issue #133, opened 2019-12-02
[request] add a `success_only` parameter to `status` command


blordpluto:
Hi guys,

Our use case is that we are doing multiple flavored deploys from a single repo. If there is a build or deploy failure, we want to message a Slack channel for the devs to fix it. `fail_only` works for that. Devs don't need to be bothered if there is no error. Great.

But if there is a successful deploy, we want to notify the testing and reliability people, in a different channel. BUT the testers don't care about failures. They only want to know when something new has successfully been deployed. And there is a lot of volume so controlling noise is important.

I have glanced at the source and honestly I could probably edit it in myself in about 5 minutes. Just an if statement with an early return, following the current pattern. But I am loathe to fork this and get into the orb business, and I suspect that there will be others out there who will need this feature just as much as I do. So I want to approach you first about getting a `success_only` parameter baked into your standard orb.

Or if there is an easy workaround I would love to know!

Thanks in advance for any help.

blordpluto:
Hi, can you please respond? For reference, I have created a branch [here](https://github.com/blordpluto/slack-orb/tree/success-only). Would like to submit a PR.

KyleTryon:
Changes requested on PR.

PR #140, opened 2020-01-27
Add a success_only parameter

blordpluto:
### Checklist

### Motivation, issues

Our use case is that we are doing multiple flavored deploys from a single repo. If there is a build or deploy failure, we want to message a Slack channel for the devs to fix it. fail_only works for that. Devs don't need to be bothered if there is no error. Great.

But if there is a successful deploy, we want to notify the testing and reliability people, in a different channel. BUT the testers don't care about failures. They only want to know when something new has successfully been deployed. And there is a lot of volume so controlling noise is important.

See my earlier request at issue https://github.com/CircleCI-Public/slack-orb/issues/133.

### Description

The new `success_only` parameter in this PR is very similar to the current `fail_only` parameter, just with the logic inverted.


KyleTryon:
I think this is a valid use-case but I would vote to remove the `notify-on-success` command. The `notify` command can handle this with the parameter as you have added, and the notify command could be called twice if desired (just supply a custom `name` parameter).

Thank you.

blordpluto:
@KyleTryon Thanks for the quick feedback. Please let me know if this change ^^^ works for you.

KyleTryon:
Much appreciated. Merging in 🎉

blordpluto:
Great! 🙌  Can I ask when we might expect a 3.4.3 release that includes this?

PR #147, opened 2020-04-09
Fix documentation

caleb15:
### Checklist

- [X] All new jobs, commands, executors, parameters have descriptions
- [X] Examples have been added for any significant new features
- [X] README has been updated, if necessary

### Motivation, issues

Incorrect doc

### Description

token doesn't work, it needs to be a full webhook url with token


caleb15:
Going by the incorrect documentation I added just the token but that resulted in the following error:

```
curl: (6) Could not resolve host: xxxxxxxxx
Job completed successfully. Alert sent.
```

By fixing the documentation people will know to use the full url, not just the token.

PR #160, opened 2020-08-13
[semver:major] Major Slack Orb Rewrite

gmemstr:
More details to come


KyleTryon:
todo: 
add a branch filter to the command.

Jobs can be filtered by branch but individual commands are not. Some users have expressed a desire to use the notify command within a job, but only notify on a given branch.

KyleTryon:
Note/Issue.

The orb is not fully POSIX compliant due to (at least) several instances where we used `[[]]` for various reasons. We should be able to address this with AWK.
https://serverfault.com/a/705144

Continuing for the moment ignoring this. We can attempt to improve and make POSIX compliant in a later step.

KyleTryon (approved):
Finally ready for launch :tada:

PR #167, opened 2020-09-03
4.0 Branch Filter.

KyleTryon:
Branch filter added.
Updated on-hold job to include all parameters.
More tests.

```yaml
description: >
  Use the "branch_list" parameter to limit notifications to specified branches.
  Useful when a job is executed on multiple branches but you only wish to be notified on a subset of branches.
  For example: If your "build" job runs on every branch, but you wish to only be notified when a failure occurs on the "staging" branch, your config may look like this.
usage:
  version: 2.1
  orbs:
    slack: circleci/slack@4.0
  jobs:
    build:
      machine: true
      steps:
        - run: some build command
        - slack/notify:
            branch_list: main # a comma separated list of branches for which this notification will execute. By default, all branches will trigger.
            event: fail
            template: basic_fail_1
        - slack/notify:
            branch_list: production # A second alert to ping certain folks to failures on this branch.
            event: fail
            template: basic_fail_1
            mentions: "<@U8XXXXXXX>, @UserName"
  workflows:
    deploy_and_notify:
      jobs:
        - build
        # This deploy job uses the native CircleCI workflow filters to specify that the entire job will not execute unless on the specified filtered branch.
        # This Workflow level filter is independent from the Slack orb.
        - deploy:
            requires:
              - build
            filters:
              branches:
                only:
                  - production

```


src/commands/notify.yml
```diff
@@ -24,6 +24,11 @@ parameters:
     type: enum
     enum: ["fail", "pass", "always"]
     default: "always"
+  branch_list:
```
gmemstr, line 27: Might be more accurate to call this "branch_pattern"

src/scripts/notify.sh
```diff
@@ -79,6 +79,25 @@ InstallJq() {
 
 }
 
+BranchFilter() {
```
gmemstr, line 82: Doesn't appear this is actually used anywhere?

gmemstr (changes requested):

src/commands/notify.yml
KyleTryon: Fair point, I like it

KyleTryon:
Ill take a look at the error shortly.

gmemstr (approved):
LGTM!

PR #169, opened 2020-09-17
4dot0 macos testing

KyleTryon:
Support macos integration testing


KyleTryon:
pushing to do more work.

Issue #77, opened 2019-08-27
[Question / Request] Tokens / multiple channels


timothyclarke:
I'm wanting to send slack messages to different teams in different channels. 

It looks like a web hook is linked to a specific channel. Is there any way to use a token and then specify the channel?

https://get.slack.help/hc/en-gb/articles/215770388-Create-and-regenerate-API-tokens

KyleTryon:
Hello @timothyclarke At the moment we don't support multiple webhooks as a parameter but I would be willing to take that as a feature request. At the moment I believe the best way to do this would to be to call the command twice, each with the different webhook value.

kamontat:
I will try this. Maybe we can change to use apps token instead? for multiple channels in slack?

timothyclarke:
Just to give this some more context. I'm not talking about 2 channels. I'm talking closer to 10. 

As builds go into DEV environments the appropriate team needs to know. 
We're using micro services/ containers so as a build / container progress through the various environments to Production different teams should be aware. Eg QA

gmemstr:
I wonder if [contexts](https://circleci.com/docs/2.0/contexts/) would help in this scenario.

timothyclarke:
I think this one was solved with Orb version 3.4.0  (which introduced the `channel` param).  I'll test and if so will close this

KyleTryon:
So, I was looking into this specifically for the new version of the slack orb (version 4.0.0) due to be released next week, however I did not include the channel parameter because of this: 
![image](https://user-images.githubusercontent.com/33272306/93779013-7dd62680-fbf4-11ea-9868-cc9177448498.png)
https://api.slack.com/messaging/webhooks

What I would recommend is creating separate webhooks for each channel and save the env vars separately in CircleCI (which could be done with contexts if desired) and when calling the orb, you can pass the channel to the webhook parameter.

```
slack/notify:
  webhook: $SLACK_CHANNEL_1
```

timothyclarke:
@KyleTryon Thanks. One of the issues I'm trying to address is that multiple (currently 2) teams / channels want to be notified of Production releases.  
* Development teams manage their own builds / deployments and have a slack channel they want notified
* There is another channel which includes the service desk and various other managers who *need* to know

Running a running a step twice with different inputs isn't an issue, running multiple jobs just for notifications become a little tedious. 
Parsing multiple webhooks also feels wrong

timothyclarke:
[...]

PR #170, opened 2020-09-21
Switched from webhook to oauth token to allow channel customization

timothyclarke:
I've got to update the documentation

This addresses https://github.com/CircleCI-Public/slack-orb/issues/77

 I've switched the code to use the app's oauth token as there is no longer a webhook I've defaulted the channel to `#general` as as it's the only one I can guarantee would be available in slack, so channel is effectively a mandatory field. 

The code has switched but I haven't amended the documentation. As this is a breaking change I'd like to get some feedback before I send time on the documentation. 
Please let me know if you'd be likely to accept this PR (and I'll go through the documentation), or if this is not a direction you want to take (and we can close this PR). 

### Checklist

### DIFF
--- a/src/commands/approval.yml
+++ b/src/commands/approval.yml
@@ -1,10 +1,11 @@
 description: Send a notification that a manual approval job is ready
 
 parameters:
-  webhook:
-    description: Enter either your Webhook value or use the CircleCI UI to add your token under the 'SLACK_WEBHOOK' env var
+  oauth_token:
     type: string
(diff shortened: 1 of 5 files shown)


timothyclarke:
This is primarily an example in code. I'm happy to rework it off https://github.com/CircleCI-Public/slack-orb/pull/160 if you would prefer

timothyclarke:
Code should be in V4

Issue #194, opened 2020-10-16
Update Docs: `@` mention Slack App in channel required permissions


JohnMahowald:
## Describe Request:
Requesting an update to the setup wiki documentation, and potentially an update to the notify.sh script.

I recently was integrating CircleCi with the slack orb. The [setup documentation](https://github.com/CircleCI-Public/slack-orb/wiki/Setup) is missing a critical step where an installing user must *add their newly created application to the channels it will notify*.

This was particularly hard to debug, because of the `> /dev/null` at the end of this line of the `notify.sh` script.

https://github.com/CircleCI-Public/slack-orb/blob/master/src/scripts/notify.sh#L41

It may make sense to remove the `> /dev/null` the response of the `notify` curl request to help developers in the future reading the CI console output. I'll put together a PR in the event that is useful.

Thanks!

## Examples:

Here is an example of a potential addition to the documentation:
https://github.com/JohnMahowald/slack-orb/wiki/Setup

## Supporting Documentation Links:
https://github.com/CircleCI-Public/slack-orb/blob/master/src/scripts/notify.sh#L41
https://github.com/JohnMahowald/slack-orb/wiki/Setup

EDIT: created a pr - https://github.com/CircleCI-Public/slack-orb/pull/195

JohnMahowald:
@KyleTryon 👆🙏👊

gmemstr:
I feel the more correct way around this is to add the `chat:write.public` permission to the bot, rather than requiring an invite to each channel.

KyleTryon:
@JohnMahowald I would love to know if this did not work for you as intended. When creating your application, was the `chat:write` scope given to the OAuth token? 
https://github.com/CircleCI-Public/slack-orb/wiki/Setup#2-permissions

https://api.slack.com/scopes/chat:write

As a bot, I believe this should give it access to post in any non-private channel.

JohnMahowald:
@KyleTryon 

I created the application with the prescribed scopes. Here is a screenshot from the app I created:
<img width="708" alt="Slack_API__Applications___Wednesday_Slack" src="https://user-images.githubusercontent.com/4701587/96299409-e1abff00-0fa8-11eb-9e42-73198b61f2c0.png">

After some debugging, I found the response I was getting from the `chat.postMessage` api was
```
{
    "ok": false,
    "error": "not_in_channel"
}
```
I `@` mentioned my new App in the group, sent an invitation, and then was able to get a success response.

[...]

PR #195, opened 2020-10-16
[feat] include curl response in CI output [semver:patch]

JohnMahowald:
See https://github.com/CircleCI-Public/slack-orb/issues/194

I found it challenging to debug the permission issues I experienced when integrating with the `slack-orb` notify command. I believe including the `curl` output in the console, or parsing the response and showing a helpful error  in the CI console may be worth considering.

Lemme know what you think. Many thanks 🙏

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -38,7 +38,7 @@ PostToSlack() {
         curl -s -f -X POST -H 'Content-type: application/json' \
         -H "Authorization: Bearer $SLACK_ACCESS_TOKEN" \
         --data \
-        "$SLACK_MSG_BODY" https://slack.com/api/chat.postMessage > /dev/null
+        "$SLACK_MSG_BODY" https://slack.com/api/chat.postMessage | jq '{ok: .ok, error: .error}'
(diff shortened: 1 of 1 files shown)


gmemstr:
I definitely think this is worth adding, but the response from Slack can be a bit verbose - we can probably take advantage of `jq`, which is already installed, to pretty print relevant parts of the response.

JohnMahowald:
@gmemstr Thanks for the recommendation. I was feeling a little lazy and didn't want to learn how to use `jq` 🤫

Anyhow, have a look and let me know what you think. Here are the expected outputs:

Success:
```
{
  "ok": true,
  "error": null
}
```

Error:
```
{
  "ok": false,
  "error": "not_in_channel"
}
```

gmemstr:
Looks good to me - @KyleTryon would love your :eyes: regarding the formatting.

KyleTryon:
That looks great 👍  I think it's a great addition that keeps the output clean but improves the usefulness. Approving this for a patch update.

KyleTryon (approved):

Issue #187, opened 2020-10-13
Support tag filtering alongside branch filtering


alecwcp:
## Describe Request:
The ability to filter by branch is really useful - we use this to filter notifications to only be sent when a deployment to one of our staging sites (controlled by branch name) fails, and also have a separate notification for failed develop branch builds.

Our production deploys are done on a semver tag. CircleCI supports filtering the jobs themselves based on the tag or the branch, but this orb currently only supports branch filtering. Adding tag filtering too would be a useful addition to allow us to notify on these with a separate message / to a separate channel / etc.
## Examples:
Since `branch_pattern` is used for branches, a new parameter of `tag_pattern` that accepts a regex might be fitting. It would work as an OR condition with the `branch_pattern` - so one notification section could be used for both tagged builds and branch builds.
## Supporting Documentation Links:
- `branch_pattern` parameter documentation here https://circleci.com/developer/orbs/orb/circleci/slack
-  Filtering jobs by branch/tag documentation here https://circleci.com/docs/2.0/workflows/#using-regular-expressions-to-filter-tags-and-branches

KyleTryon:
Hello,

Does the following use-case solve your need?

https://circleci.com/developer/orbs/orb/circleci/slack#usage-successful_tagged_deployment

```
version: '2.1'
orbs:
  node: 'circleci/node:4.0'
  slack: circleci/slack@4.0
jobs:
  deploy:
    executor:
      name: node/default
    steps:
      - checkout
      - node/install-packages
      - run:
          command: npm run deploy
      - slack/notify:
          channel: ABCXYZ
          event: success
          template: basic_fail_1
workflows:
  deploy_and_notify:
    jobs:
      - deploy:
          context:
            - slack-secrets
```

Workflow filters for instance already support tagging, so that a job will only run in the event the job has been tagged, at which point, no special parameter for the orb will be needed.

alecwcp:
Quick response :slightly_smiling_face: 

Currently the only difference between the `deploy_prod` and `deploy_nonprod` jobs is the context. So to avoid duplication in our config file we define the deploy job once then use yaml anchors to use it for both the `deploy_prod` and `deploy_nonprod` jobs with just the context difference.

I think we could pull some of that out, it would create some duplication in the file but would accomplish what you suggest above.

[...]

PR #196, opened 2020-10-16
Filter patterns by current branch and/or current tag [semver:minor]

### LINKED ISSUE #192: No notification is sent with empty branch name
## Orb version:

4.0.1

## What happened:

On tag push release, branch name is set empty; default regex pattern /.+/  does not match and no notification is sent.

Preparing environment variables

```
Using build environment variables:
  BASH_ENV=/tmp/.bash_env-5f872c9e54f5364a376525ef-0-build
  CI=true
  CIRCLECI=true
  CIRCLE_BRANCH=  # EMPTY HERE
  CIRCLE_BUILD_NUM=674
   ...
  CIRCLE_STAGE=release
  CIRCLE_TAG=v1.0.0
  ...
```

Final message

```
NO SLACK ALERT

Current branch does not match any item from the "branch_pattern" parameter
Current branch: 
CircleCI received exit code 0
```

## Expected behavior:

Notification should always be sent, even if branch is empty.
I believe the default `branch_pattern` should be `.*` instead of `.+`

## Additional Information:

https://github.com/CircleCI-Public/slack-orb/blob/v4.0.1/src/commands/notify.yml#L23-L27

skyeagle:
Added a filter pattern by the current tag.

Closes #192 and #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,13 +52,12 @@ The Slack Orb comes with a number of included templates to get your started with
         }
   ```
 
+## Branch or Tag Filtering
 
-## Branch Filtering
-
-Limit Slack notifications to particular branches with the "branch_pattern" parameter.
+Limit Slack notifications to particular branches with the "branch_pattern" or "tag_pattern" parameter.
 
 ```
-A comma separated list of regex matchable branch names. Notifications will only be sent if sent from a job from these branches. By default ".+" will be used to match all branches. Pattern must match the full string, no partial matches.
+A comma separated list of regex matchable branch or tag names. Notifications will only be sent if sent from a job from these branches/tags. By default ".+" will be used to match all branches/tags. Pattern must match the full string, no partial matches. Keep in mind that "branch_pattern" and "tag_pattern" are mutually exclusive.
 ```
 
 See [usage examples](https://circleci.com/developer/orbs/orb/circleci/slack#usage-examples).
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -25,6 +25,11 @@ parameters:
       A comma separated list of regex matchable branch names. Notifications will only be sent if sent from a job from these branches. By default ".+" will be used to match all branches. Pattern must match the full string, no partial matches.
     type: string
     default: ".+"
+  tag_pattern:
+    description: |
+      A comma separated list of regex matchable tag names. Notifications will only be sent if sent from a job from these branches. By default ".+" will be used to match all tags. Pattern must match the full string, no partial matches.
+    type: string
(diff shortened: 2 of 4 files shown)


gmemstr (dismissed):
Debating whether to mark this as a minor or patch semver. Looks good otherwise!

KyleTryon:
Wondering if we should break out tag into its own parameter as mentioned here: https://github.com/CircleCI-Public/slack-orb/issues/187

skyeagle:
Is there a particular reason for having a separate parameter for tag_pattern? Looks like the regexp condition is flexible enough. What do you guys think?

BeyondEvil:
+1 on having a separate parameter for tags, and the parameter should be regex enabled (not just true/false).

For reusability reasons, there can be multiple workflows with different tag-patterns using the same job/command.

skyeagle:
I've updated it with a separate pattern for the current tag. Thanks!

BeyondEvil:
Since they are mutually exclusive, is it worth having a check for that? Or at least put it in the description? Save a newbie some headache in the future. 😂

skyeagle:
good point @BeyondEvil, I've updated the readme. thanks for the review!

KyleTryon:
For posix compliance (to make the alpine users happy), could we swap out this bash if statement with the shell equivalent?
https://app.circleci.com/pipelines/github/CircleCI-Public/slack-orb/690/workflows/32aae1cc-acbd-4bfd-9b1a-ec0892a34d69/jobs/2475

BASH

```
if [[ -z $1 || -z $2 ]]
```

SHELL

```
if [ -z $1 ] || [ -z $2 ]
```

I want to run some additional tests here manually as well. Im not sure for instance, but I believe currently the branch filter might also match no branch at all, like there would be if it was a tagged commit.

skyeagle:
@KyleTryon thanks for pointing out, I've made suggested changes.

BeyondEvil:
> I want to run some additional tests here manually as well. Im not sure for instance, but I believe currently the branch filter might also match no branch at all, like there would be if it was a tagged commit.

I tried that, but it doesn't match no branch (empty string). But please try it yourself.

skyeagle:
any chance we can merge this? It's a real showstopper for my team.

gmemstr:
I feel comfortable with the PR right now, but I'll let @KyleTryon have the final say :)

KyleTryon:
Merging after #202. 202 is a patch, this release will go out as a minor (for the new feature)

KyleTryon (approved):
Looking good 👍  releasing as a minor update.

PR #198, opened 2020-10-22
Fixes #186: Escapes apostrophe in custom body

### LINKED ISSUE #186: Crashes on apostrophe
## Orb version:

4.0.0
## What happened:

Orb crashes when the payload contains an apostrophe
## Expected behavior:
Doesn't crash

## Additional Information:
Config:

```yml
      - slack/notify:
          channel: test-webhook
          custom: |
            {
              "blocks": [
                {
                  "type": "section",
                  "text": {
                    "type": "mrkdwn",
                    "text": "I'm Circle CI"
                  }
                }
              ]
            }
          event: always
```

Orb output:

```
/bin/bash: eval: line 8: unexpected EOF while looking for matching `''
/bin/bash: eval: line 9: syntax error: unexpected end of file

Exited with code exit status 1

CircleCI received exit code 1
```

If I change the text to `"Im Circle CI"`, it doesn't crash.

n-thumann:
Fixes https://github.com/CircleCI-Public/slack-orb/issues/186 by escaping `'` properly :v:

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -5,7 +5,7 @@ BuildMessageBody() {
     #   If none, error.
     if [ -n "$SLACK_PARAM_CUSTOM" ]; then
         ModifyCustomTemplate
-        CUSTOM_BODY_MODIFIED=$(echo "$CUSTOM_BODY_MODIFIED" | sed 's/"/\\"/g' | sed 's/\\n/\\\\n/g' | sed 's/|/\\|/g' | sed 's/</\\</g' | sed 's/>/\\>/g')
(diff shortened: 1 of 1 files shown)


n-thumann:
I am struggling to reproduce the failed CI step: `bats ./src/tests` runs fine for me, can someone check or trigger a rerun?

KyleTryon:
It will require further digging to see why this failed. I have rerun the job with the same result. I assume the output somehow must not be matching. 

I was looking into this previously and was given this code by @ryanwohara I had hoped to bake in as a new function in the code that would be separately testable. A function named `SanitizeInput` or similar. I have not yet however had the chance to play with it.

```
echo "'{}\\" | sed "s|\(['\{}]\)|\\\\\1|g"

\'\{\}\\
``` 

This should automatically escape the characters within the `[]` with their escaped counterparts. (I think we may need to actually double escape).

KyleTryon:
Thank you @n-thumann. We are currently reviewing a separate PR that seems to resolve this issue in a different way that may come with additional benefits: https://github.com/CircleCI-Public/slack-orb/pull/199

For this reason, we are going to close this PR, but we thank you for your contribution 🙏

PR #199, opened 2020-10-23
[semver:patch] double quote inputs inside eval

### LINKED ISSUE #182: Backticks for inline code formatting are incorrectly parsed
## Orb version: 
4.0.0

## What happened: 
Backticks for inline code formatting hide the content inside them. If I try to escape them with `\`, build fails with message: `parse error: Invalid escape at line 1, column 291`

## Expected behavior:

Backtics should add inline-code formatting to string as it is done in slack app.

### LINKED ISSUE #186: Crashes on apostrophe

ark-1:
Messages with parentheses were failing the step, and asterisk sometimes was interpreted as a part of glob pattern.

It seems there is an endless list of things that need to be escaped (#182, #186 and also `(`, `)`, `*`, `{`, etc.), because inside `eval`, the echo command accepted unquoted arguments. Now there's quoting, so a lot of things that were previously escaped don't need to be escaped now.
Another pleasant side effect is that repeated whitespace in custom template message text is now preserved.

Also added a test for this behavior.

Closes #186, closes #182


gmemstr:
@ark-1 Do you forsee this breaking any existing templates? I'm happy with this being a patch, but want to verify :)

gmemstr (approved):

ark-1:
No, they souldn't. The only thing that could be used (but I doubt it) but now doesn't work is command substitution via back-quotes in templates:
``"current date and time: `date +'%m/%d/%Y %T'` "``
I believe the ability to use back-quotes to format code in markdown messages is more important, and users should use (and most probably are using) `$(cmd)` for command substitution

KyleTryon:
shipping time

Thank you @ark-1 ! This looks to be a brilliant solution and the tests notifications look great 👍 . Shipping as a patch now.

PR #202, opened 2020-10-26
[semver:patch] Usage example update

### LINKED ISSUE #185: Should throw error on missing SLACK_ACCESS_TOKEN
## Orb version:

4.0.0

## What happened:

Notify job passes but no notification is sent.

## Expected behavior:

Notification sent successfully.

## Additional Information:

I had forgotten to set the SLACK_ACCESS_TOKEN environment variable. Was quite surprised when I realized that the issue was the missing env var and no error had been produced.

### LINKED ISSUE #200: Unclear event values
In examples provided in https://circleci.com/developer/orbs/orb/circleci/slack#usage-successful_tagged_deployment,

event key is associated to 'success, while the Orb description indicates values can be one of 'pass', 'fail' and 'always'.

Which one is correct between 'pass' and 'success?'

KyleTryon:
Fixes #200 - Incorrect parameter value for event.
Fixes #185 - Improved error messaging for missing required values.

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -113,7 +113,7 @@ CheckEnvVars() {
         exit 1
     fi
     # If no channel is provided, quit with error
-    if [ "$SLACK_PARAM_CHANNEL" = "" ]; then
+    if [ -z "$SLACK_PARAM_CHANNEL" ]; then
        echo "No channel was provided. Enter value for SLACK_DEFAULT_CHANNEL env var, or channel parameter"
        exit 1
     fi
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -128,7 +128,6 @@ CheckEnvVars() {
 # This is done so this script may be tested.
 ORB_TEST_ENV="bats-core"
 if [ "${0#*$ORB_TEST_ENV}" = "$0" ]; then
-    . "/tmp/SLACK_JOB_STATUS"
     CheckEnvVars
     InstallJq
     BuildMessageBody
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -27,11 +27,6 @@ PostToSlack() {
     # Post once per channel listed by the channel parameter
     #    The channel must be modified in SLACK_MSG_BODY
 
-    # If no channel is provided, quit with error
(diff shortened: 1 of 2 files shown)


KyleTryon:
Looking for other tiny items to add to this patch.

KyleTryon:
Fixes #185

src/scripts/notify.sh
```diff
@@ -111,11 +106,29 @@ BranchFilter() {
...
+        echo "Please note, Webhooks are no longer used for the Slack Orb (v4 +)."
+        echo "Follow the setup guide available in the wiki: https://github.com/CircleCI-Public/slack-orb/wiki/Setup"
+    fi
+}
+
 # Will not run if sourced from another script.
 # This is done so this script may be tested.
 ORB_TEST_ENV="bats-core"
 if [ "${0#*$ORB_TEST_ENV}" = "$0" ]; then
-    . "/tmp/SLACK_JOB_STATUS"
```
KyleTryon, line 118: CC @gmemstr I think this may be from the old orb as the blame shows you added this, but it was during the major re-write. Im testing now removing this. Do you have any idea what this might have been? Looks to be sourcing a tmp file that I wouldnt expect to exist. I think it was possibly used in the old orb, not sure how I didnt see it.

gmemstr: I don't recall removing/adding `. "/tmp/SLACK_JOB_STATUS"`, so I don't really know :(

gmemstr (approved):

src/scripts/notify.sh
skyeagle: Looks like this has been referenced for carrying over the job's status and removing this introduced a bug.

I see this output, but not sure where it comes from:

*Slack - Detecting Job Status (FAIL)*
```
echo 'export CCI_STATUS="fail"' > /tmp/SLACK_JOB_STATUS
```

Specifying `event: fail` stopped working for us.

KyleTryon: Ahh it was hidden in the other command / run step of course 🤦 very sorry about that. Patching this now

KyleTryon: Thank you @skyeagle

Issue #212, opened 2020-11-18
Notify command should stop as soon as possible when no notification should be sent


ralbertazzi:
The line referenced below should be executed as soon as possible so we avoid taking unnecessary steps (such as installing dependencies) if no notification should be sent (e.g. the notification should be sent on fail but the job passed). This will make the CI faster when no Slack message needs to be sent.

https://github.com/CircleCI-Public/slack-orb/blob/5435b8129cdade77c990e6e07ae7de0f0a4c0261/src/scripts/notify.sh#L43

gmemstr:
Fixed in #213

ralbertazzi:
Thank you!! 🙏

PR #213, opened 2020-11-23
[semver:patch] Decide earlier whether to post message

gmemstr:
See #212

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -137,7 +137,7 @@ ORB_TEST_ENV="bats-core"
 if [ "${0#*$ORB_TEST_ENV}" = "$0" ]; then
     ShouldPost
     CheckEnvVars
-    source "/tmp/SLACK_JOB_STATUS"
+    . "/tmp/SLACK_JOB_STATUS"
     InstallJq
     BuildMessageBody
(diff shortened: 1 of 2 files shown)


src/scripts/notify.sh
```diff
@@ -133,13 +113,33 @@ CheckEnvVars() {
...
+}
+
 # Will not run if sourced from another script.
 # This is done so this script may be tested.
 ORB_TEST_ENV="bats-core"
 if [ "${0#*$ORB_TEST_ENV}" = "$0" ]; then
-    . "/tmp/SLACK_JOB_STATUS" # this is required for fetching the job status set by the notify.yml command.
+    ShouldPost
     CheckEnvVars
+    . "/tmp/SLACK_JOB_STATUS"
```
KyleTryon, line 140: While we are here, we should change the `.` to `source` for POSIX

KyleTryon (approved):

KyleTryon (approved):

src/scripts/notify.sh
nickpalmer, line 140: This is the bug in 4.1.2. This has to come before ShouldPost since that uses CCI_STATUS to decide if the job should be run.

PR #220, opened 2020-12-22
[semver:minor] New Feature: Exit_On_Fail

KyleTryon:
Changes:

++ New parameter `ignore_errors` of type `boolean`
Description:
> Ignore errors posting to Slack.
> Disable to catch initial setup errors. Re-enable to prevent Slack errors from affecting your pipeline.

### DIFF
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -42,12 +42,12 @@ parameters:
       Select which channel in which to post to. Channel name or ID will work. You may include a comma separated list of channels if you wish to post to multiple channels at once. Set the "SLACK_DEFAULT_CHANNEL" environment variable for the default channel.
     type: string
     default: $SLACK_DEFAULT_CHANNEL
-  exit_on_fail:
+  ignore_errors:
       description: |
-        Fail the pipeline if there has been an error posting to Slack.
-        This setting is useful mostly for initial setup to ensure the orb and Slack have been configured properly. Once established, it is recommended to disable this setting to ensure your pipelines are not inhibited by potential notification issues.
+        Ignore errors posting to Slack.
+        Enable this to prevent Slack outages from affecting your pipeline.
       type: boolean
-      default: true
+      default: false
 steps:
   - run:
       when: on_fail
@@ -70,7 +70,7 @@ steps:
         SLACK_PARAM_BRANCHPATTERN: "<<parameters.branch_pattern>>"
         SLACK_PARAM_TAGPATTERN: "<<parameters.tag_pattern>>"
         SLACK_PARAM_CHANNEL: "<<parameters.channel>>"
-        SLACK_PARAM_EXIT_ON_FAIL: "<<parameters.exit_on_fail>>"
+        SLACK_PARAM_IGNORE_ERRORS: "<<parameters.ignore_errors>>"
         # import pre-built templates using the orb-pack local script include.
         basic_fail_1: "<<include(message_templates/basic_fail_1.json)>>"
         success_tagged_deploy_1: "<<include(message_templates/success_tagged_deploy_1.json)>>"
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -42,7 +42,12 @@ parameters:
       Select which channel in which to post to. Channel name or ID will work. You may include a comma separated list of channels if you wish to post to multiple channels at once. Set the "SLACK_DEFAULT_CHANNEL" environment variable for the default channel.
     type: string
     default: $SLACK_DEFAULT_CHANNEL
-
+  exit_on_fail:
+      description: |
+        Fail the pipeline if there has been an error posting to Slack.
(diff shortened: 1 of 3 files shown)


src/scripts/notify.sh
```diff
@@ -32,10 +32,17 @@ PostToSlack() {
...
         SLACK_MSG_BODY=$(echo "$SLACK_MSG_BODY" | jq --arg channel "$i" '.channel = $channel')
-        curl -s -f -X POST -H 'Content-type: application/json' \
-        -H "Authorization: Bearer $SLACK_ACCESS_TOKEN" \
-        --data \
-        "$SLACK_MSG_BODY" https://slack.com/api/chat.postMessage | jq '{ok: .ok, error: .error}'
+        SLACK_SENT_RESPONSE=$(curl -s -f -X POST -H 'Content-type: application/json' -H "Authorization: Bearer $SLACK_ACCESS_TOKEN" --data "$SLACK_MSG_BODY" https://slack.com/api/chat.postMessage)
+        SLACK_ERROR_MSG=$(echo "$SLACK_SENT_RESPONSE" | jq '.error')
+        if [ ! "$SLACK_ERROR_MSG" = "null" ]; then
+            echo "There was an error sending the Slack message. Please view the error below."
+            echo "ERROR: $SLACK_ERROR_MSG"
```
alexjurkiewicz, line 39: It might be good to remove the `ERROR: ` prefix here, to reduce ambiguity about where the Slack API response begins and ends.

```sh
echo "Slack API returned an error message:"
echo "$SLACK_ERROR_MSG"
```

src/commands/notify.yml
```diff
@@ -42,7 +42,12 @@ parameters:
       Select which channel in which to post to. Channel name or ID will work. You may include a comma separated list of channels if you wish to post to multiple channels at once. Set the "SLACK_DEFAULT_CHANNEL" environment variable for the default channel.
     type: string
     default: $SLACK_DEFAULT_CHANNEL
-
+  exit_on_fail:
```
alexjurkiewicz, line 45: This name is a bit weird. The command will always exit whether the API request succeeds or fails. Suggestions:

* `ignore_failures` (inverted condition)
* `report_failures`

src/commands/notify.yml
```diff
@@ -42,7 +42,12 @@ parameters:
...
+      description: |
+        Fail the pipeline if there has been an error posting to Slack.
+        This setting is useful mostly for initial setup to ensure the orb and Slack have been configured properly. Once established, it is recommended to disable this setting to ensure your pipelines are not inhibited by potential notification issues.
```
alexjurkiewicz, line 48: This could be a little more terse, and use present tense.

```suggestion
      description: |
        Fail on errors posting to Slack. Disable this to prevent Slack outages affecting your pipeline.
```

src/commands/notify.yml
KyleTryon: Excellent suggestion

src/commands/notify.yml
KyleTryon: Again, excellent suggestion. Thank you for your input 👍. Applying all of these now.

src/commands/notify.yml
KyleTryon: @alexjurkiewicz How would you feel about `ignore_errors` as to not potentially confuse `failures` with the `fail` event and CI job failures?

alexjurkiewicz: good point. Sounds perfect 👌

KyleTryon:
One final thought (hopefully). It may not be proper to issue this as a minor release in its current state, and may require modification of the default value (losing most/all benefit), or potentially looking at a new major version.... or maybe breaking semver slightly.

**Issue:** 
This new parameter and feature, adds a toggle to ignore errors posting to slack, with the default set to false, which will result in a pipeline error if the orb can not post to slack. This is useful for initial setup to catch config and permission errors. Once setup is complete, the setting should be switched to true, to prevent slack errors from stopping a pipeline.

However, adding this feature changes the "default" of how the orb works today, which is to always ignore all errors. So this change will certainly result in some users utilizing version `4` failing. Also users who are on `4.1` and later update to this `4.2` version.

The _proper_ solution would be to issue version `5.0.0` for this change. I would like to avoid this if possible but opening the discussion.

KyleTryon:
Going on a suggestion from @felicianotech, I think we may need to change the default so that we may merge in as a minor change, and take a note to create a milestone to change this default in the next major update.

felicianotech:
@KyleTryon Just took a look. I did a review but before I give the green checkmark, I see that the default is set to true. Weren't you going to change that or did I misunderstand?

KyleTryon:
> @KyleTryon Just took a look. I did a review but before I give the green checkmark, I see that the default is set to true. Weren't you going to change that or did I misunderstand?

This is correct, it was changed to true in the last commit in response to your suggestion. Setting this option to true by default will ignore errors, as the orb currently does. The desired initial state is false but would break semver.

felicianotech (approved):

felicianotech:
So I did misundertand.. the initial state. Cool, thanks. This looks good (duh, I already reviewed).

PR #226, opened 2021-01-05
[semver:minor] Create hook to inject variables into messages

arigesher:
Place a hook into the `notify.sh` script that will source `/tmp/slack-notify-enhance` to allow additional variables to be used in the build notification messages

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -139,6 +139,12 @@ ShouldPost() {
     fi
 }
 
+# allow the injection of other data into the environment
+# by the build to be used in slack messages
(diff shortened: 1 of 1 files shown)


gmemstr (changes requested):
I think it's probably worth documenting this somewhere, preferably in an example (`src/examples`).

KyleTryon:
The user can already store any environment variables they wish in $BASH_ENV and it will be automatically sourced by the orb.
https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-shell-command
(for bash)

KyleTryon:
Closing due to lack of comments. Please feel free to comment to re-open.

PR #228, opened 2021-01-13
Add basic_success_1.json template

montanaflynn:
Matches other basic templates.

The existing `success_tagged_deploy_1.json` template isn't applicable to my workflows and shows an empty tags section.

### DIFF
--- a/src/message_templates/basic_success_1.json
+++ b/src/message_templates/basic_success_1.json
@@ -1,0 +1,60 @@
+{
+	"blocks": [
+		{
+			"type": "header",
+			"text": {
+				"type": "plain_text",
+				"text": "Job Succeeded. :green_circle:",
+				"emoji": true
+			}
+		},
+		{
+			"type": "section",
+			"fields": [
+				{
+					"type": "mrkdwn",
+					"text": "*Job*: ${CIRCLE_JOB}"
+				}
+			]
+		},
+		{
+			"type": "section",
+			"fields": [
+				{
+					"type": "mrkdwn",
+					"text": "*Project*:\\n$CIRCLE_PROJECT_REPONAME"
+				},
+				{
+					"type": "mrkdwn",
+					"text": "*Branch*:\\n$CIRCLE_BRANCH"
+                },
+                {
+					"type": "mrkdwn",
+					"text": "*Commit*:\\n$CIRCLE_SHA1"
+				},
+				{
+					"type": "mrkdwn",
+					"text": "*Author*:\\n$CIRCLE_USERNAME"
+				}
+			],
+			"accessory": {
+				"type": "image",
+				"image_url": "https://assets.brandfolder.com/otz5mn-bw4j2w-6jzqo8/original/circle-logo-badge-black.png",
+				"alt_text": "CircleCI logo"
+			}
+		},
+		{
+			"type": "actions",
+			"elements": [
+				{
+					"type": "button",
+					"text": {
+						"type": "plain_text",
+						"text": "View Job"
+					},
+					"url": "${CIRCLE_BUILD_URL}"
+				}
+			]
+		}
+	]
+}


KyleTryon:
Hello @montanaflynn thank you for the PR. Could I ask, I understand you are looking to add this template as the current success template has a space for your "tag" which is irrelevant to your needs, is that right? 

I am attempting to limit the total number of templates included in the orb (especially given removing them in the future will be difficult) and so just want to ensure each available template provides maximum benefit.

Could you tell us for instance if the empty tag field causes any confusion or is visually unappealing? Would, for instance, filling in the blank space with "no tag" or something similar be acceptable?

In either case I will request some feedback from the team/community on this addition.

montanaflynn:
The main reason isn't so much the empty tag section but the lack of branch name and commit SHA which are much more useful to see in the slack message than tag name. Also the naming doesn't seem standard, since `basic_fail_1` and `basic_on_hold_1` exists this follows that pattern.

gmemstr:
I think this is a good addition, but I don't know if it has to be a completely different template - I wonder if we can address these shortcomings within the existing template.

We're a little template-adding phobic given each template we add increases the size of the orb substantially (we should definitely look into minifying these, as a sidenote @KyleTryon), and we do prefer more generic templates, hence the suggestion to try and roll this into the existing template.

KyleTryon:
> I think this is a good addition, but I don't know if it has to be a completely different template - I wonder if we can address these shortcomings within the existing template.
> 
> We're a little template-adding phobic given each template we add increases the size of the orb substantially (we should definitely look into minifying these, as a sidenote @KyleTryon), and we do prefer more generic templates, hence the suggestion to try and roll this into the existing template.

@gmemstr minifying is a great idea. @montanaflynn I agree we likely should have stuck with the basic naming convention rather than focusing on the tagged portion with is a bit opinionated around tag-based deployments. We could rename this, but it would need to wait for a major version change.

As @gmemstr has suggested, I would be open to making the existing template a more robust set of fields, including branch and SHA as @montanaflynn mentioned.

montanaflynn:
> We're a little template-adding phobic given each template we add increases the size of the orb substantially...

KyleTryon:
> > We're a little template-adding phobic given each template we add increases the size of the orb substantially...
> 
> Is this really the main concern? The JSON template is 1kb.
> 
> Having to find and use a template called `success_tagged_deploy_1` when you don't use tags is odd to me.
> 
> I think that having three "basic" templates makes the most sense, but it's your repo so you can decide what you want, but now my circleci config file went from clean yaml steps to having this giant JSON blob in it.

I think your point is fair. Give us a little bit more time to converse on this.

NewAlexandria:
I wandered in here looking to configure things like templates, and realizing that I should probably just fork this repo and deploy our own orb instead of engage.   Do hope you guys start to come around to thinking about how teams use your tools.  Thanks for offering the orb source.

KyleTryon:
> I wandered in here looking to configure things like templates, and realizing that I should probably just fork this repo and deploy our own orb instead of engage. Do hope you guys start to come around to thinking about how teams use your tools. Thanks for offering the orb source.

We encourage forking this orb. We also of course value the input of the users of this orb which is why this PR is still open, we are glad to receive feedback from other users such as yourself.

Thank you for your feedback, this will influence our decisions.

montanaflynn:
How can you easily fork the orb and use it? 

I couldn't find any information other than [publishing an orb](https://circleci.com/docs/2.0/creating-orbs/) which is a process I don't want to do. 

Honestly I really miss the old non-orb slack integration, all our older repos have a cleaner circleci config and the slack notifications are perfect IMO, small with a lot of information packed in:

Screen Shot 2021-01-15 at 08 41 54

Compared to:

Screen Shot 2021-01-15 at 08 44 46

KyleTryon:
> How can you easily fork the orb and use it?
> 
> I couldn't find any information other than [publishing an orb](https://circleci.com/docs/2.0/creating-orbs/) which is a process I don't want to do.
> 
> Honestly I really miss the old non-orb slack integration, all our older repos have a cleaner circleci config and the slack notifications are perfect IMO, small with a lot of information packed in:
> 
> Screen Shot 2021-01-15 at 08 41 54
> 
> Compared to:
> 
> Screen Shot 2021-01-15 at 08 44 46

I believe the easiest option is to use the existing orb and add your own customized template for use.
https://github.com/CircleCI-Public/slack-orb#custom-message-template
If this does not appear to be a good solution for you, I would love any feedback as to why this is unfavorable and what would be preferred.

As for non-orb-based Slack integration, that is my final goal. There will actually be less customizability with a bot, meaning you would _not_ be able to choose or customize templates. But A bot will allow us to include other features that are attractive as well, such as potentially allowing you to approve a workflow from slack, but ill save this conversation for elsewhere.

If you would like to fork the orb itself there are a few ways to do that.

1. Fork the actual repository.

If you fork or clone the project  as-is you will need to modify the configuration file to use, and publish to your own namespace, as well as add your personal API token to a context on CircleCI.

2. Copy the "packed" orb source.

The quickest method would be to simply republish the packed source under your own namespace. This is however more difficult (in my opinion) to develop for or modify as all of the code is in a single yaml file.

Get the packed source by running this command: `circleci orb source circleci/slack > my-orb-source.yml`

montanaflynn:
> I believe the easiest option is to use the existing orb and add your own customized template for use.
> https://github.com/CircleCI-Public/slack-orb#custom-message-template
> If this does not appear to be a good solution for you, I would love any feedback as to why this is unfavorable and what would be preferred.

This is what I'm currently doing but my circleci config file went from clean yaml steps to having giant JSON blobs in it. This is compounded when you have more jobs / slack notifications in a single yaml file. 

I just want a basic success message at the end of the entire workflow or a notification which job failed. This is [not possible](https://github.com/CircleCI-Public/slack-orb/issues/231) to do with the current slack orb. It was done really well with the previous `SLACK_WEBHOOK` environment variable integration that apparently isn't allowed on new circleci repos.

aguynamedben:
This is so complicated to setup compared to the old integration. There should for sure be a template like this.

Sure I can fork this repo, or add my own crazy Slack config, but a simple Slack message saying "the tests passed" should be included in this package.... 🤷‍♂️ Why make it hard on your users?

Honestly my first go-round I expected the `basic_fail_1` template do "do the right thing" if everything passed.

```yml
      - slack/notify:
          channel: C01N4B43AP2
          event: always
          template: basic_fail_1
```

But this make a fail message show up even when the build passes.

I'm using CI for just running tests, no deploying, so the `success_tagged_deploy_1` template is over-specific.

+100 there should be `basic_success_1`. There's a lot of backward reasoning that's unfriendly to the users going on here, especially given how simple and effective the old Slack integration was.

KyleTryon:
> I'm using CI for just running tests, not deploying, so the `success_tagged_deploy_1` template is over-specific. 

I think this is a valid point. We need to add a basic success template with minimal output to remove confusion. 

---

> Honestly my first go-round I expected the `basic_fail_1` template do "do the right thing" if everything passed.
> ```yaml
>       - slack/notify:
>           channel: C01N4B43AP2
>           event: always
>           template: basic_fail_1
> ```
> 
> But this makes a fail message show up even when the build passes.

The `template` is only the message content we are referring to. 

| Parameter | Description
|---|---|
| template |  Use one of the built-in templates. Select the template name. Preview each template on the GitHub Readme|
| event | In what event should this message send? Options: ["fail", "pass", "always"]|

Because `always` is selected here, we are saying to send the template in all events. Templates come from Slack: https://app.slack.com/block-kit-builder/

You can use the above link to also create your own custom slack template if the ones provided are not sufficient.

You can enter the JSON directly in your config or (CC @montanaflynn ) you could save the templates to a context ,project environment variable, or even load them from a file dynamically for instance and store them this way.

https://github.com/CircleCI-Public/slack-orb/wiki/Dynamic-Templates#2-export-template-to-environment-variable

Example custom template:
```yaml
- slack/notify:
    event: fail
    template: MY_ORB_TEMPLATE
    channel: ABCXYZ
```

From this conversation, I can see we clearly need to add a basic template for successful jobs not based on the tag. I will be testing this PR and merging shortly given no issues.

montanaflynn:
Honestly I just wish the old slack integration came back. It worked perfectly for my needs, gave a lot of information in a compact manner and "just worked" when it comes to doing a single alert at the end of a circleci job. The new slack orb integration has so many caveats and problems it's really been a headache, all our old repos have very simple slack integration that is much better than the new ones.

KyleTryon:
Hello everyone,

The changes from this PR have been included in #239,. Thank you @montanaflynn and everyone who commented.
[slack@4.3.0](https://circleci.com/developer/orbs/orb/circleci/slack?version=4.3.0) is live now.

The Docs and Readme have been updated and as a result of the conversations above I have also added a new guide for utilizing Contexts to store many Slack templates that can be reused across multiple projects and will not add any JSON to your config: https://github.com/CircleCI-Public/slack-orb/wiki/Reuse-Custom-Templates

PR #229, opened 2021-01-25
Accept params for icon, username, and color

NewAlexandria:
Hoping this seems like an acceptable feature.  

We really wanted to customize the readability based on context, without having a ton config littered around.  

Hope you'll find this is acceptable.  Revision requests welcome.

### DIFF
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -48,6 +48,21 @@ parameters:
         Disable to catch initial setup errors. Re-enable to prevent Slack errors from affecting your pipeline.
       type: boolean
       default: true
+  icon:
+    description: |
+      URL for the icon to be displayed with the post
+    type: string
+    default: ""
+  username:
+    description: |
+      The name of the poster, be displayed with the post
+    type: string
+    default: ""
+  color:
+    description: |
+      Vertical bar color on the leading-edge of the post
+    type: string
+    default: ""
 steps:
   - run:
       when: on_fail
@@ -71,6 +86,9 @@ steps:
         SLACK_PARAM_TAGPATTERN: "<<parameters.tag_pattern>>"
         SLACK_PARAM_CHANNEL: "<<parameters.channel>>"
         SLACK_PARAM_IGNORE_ERRORS: "<<parameters.ignore_errors>>"
+        SLACK_PARAM_ICON: "<<parameters.icon>>"
+        SLACK_PARAM_USERNAME: "<<parameters.username>>"
+        SLACK_PARAM_COLOR: "<<parameters.color>>"
         # import pre-built templates using the orb-pack local script include.
         basic_fail_1: "<<include(message_templates/basic_fail_1.json)>>"
         success_tagged_deploy_1: "<<include(message_templates/success_tagged_deploy_1.json)>>"


KyleTryon:
Hello @NewAlexandria,

This is very similar to the Slack orb prior to version 4, where each individual customizable part of a message could be configured via parameters. We have attempted to limit the number of parameters since 4.0 in favor of allowing the user to customize their template entirely.

The options you have provided here are saved as environment variables for the job, but they don't appear to be utilized anywhere. Is the goal for this change to make the settings available for custom templates? Or might the existing templates need to be modified to properly use this information?

If you look at Slack Orb version 3.4.2 you can see the old design (you could still use this version today if desired) where we had many more parameters such as you provided.
https://circleci.com/developer/orbs/orb/circleci/slack?version=3.4.2#commands-notify

You can now, (in version 4+) include your own completely custom templates: https://github.com/CircleCI-Public/slack-orb/wiki#custom-message-template

src/commands/notify.yml
```diff
@@ -48,6 +48,21 @@ parameters:
         Disable to catch initial setup errors. Re-enable to prevent Slack errors from affecting your pipeline.
       type: boolean
       default: true
+  icon:
+    description: |
+      URL for the icon to be displayed with the post
+    type: string
+    default: ""
+  username:
```
KyleTryon, line 56: This would indicate a static value set in the config. What would "poster" relate to? The commit author? We may have an existing environment variable for the intended user captured here

src/commands/notify.yml
```diff
@@ -48,6 +48,21 @@ parameters:
         Disable to catch initial setup errors. Re-enable to prevent Slack errors from affecting your pipeline.
       type: boolean
       default: true
+  icon:
```
KyleTryon, line 51: Followed up in my initial comment. The icon seen in our provided templates are a part of the _template_, we do not inject icons into messages. For custom icons/messages I would currently recommend using a custom template.

What would be expected if a user set an icon parameter as well as a custom template?

src/commands/notify.yml
```diff
@@ -71,6 +86,9 @@ steps:
         SLACK_PARAM_TAGPATTERN: "<<parameters.tag_pattern>>"
         SLACK_PARAM_CHANNEL: "<<parameters.channel>>"
         SLACK_PARAM_IGNORE_ERRORS: "<<parameters.ignore_errors>>"
+        SLACK_ICON: "<<parameters.icon>>"
```
KyleTryon, line 89: Orb job level env vars should be prefixed with the same pattern as the other environment variables, prefixed with `SLACK_PARAM_`

example: `SLACK_ICON` -> `SLACK_PARAM_ICON`

src/commands/notify.yml
NewAlexandria: I guess I could have said "the username that should author the post", or something similar. 

I'm open to the reality that I couldn't read the docs correctly and jumped to a PR that is party or wholly unneeded. It's no problem if that's the case.

src/commands/notify.yml
NewAlexandria: Trying to avoid custom templates since they need to be defined inline, which makes things less readable.  Maybe I can make the template assigned in an ENV var, so that it's abstracted to a file that I `cat` during a `run` step?

I also don't see anything that suggests a template lets me change the equiv of `SLACK_PARAM_ICON`, unless you mean the `accessory` example?

src/commands/notify.yml
KyleTryon: The user that authors the post is the "bot" application that you register upon setup. I do not believe this can be adjusted from the message template.

src/commands/notify.yml
KyleTryon: You absolutely can 👍 , in fact, I don't believe we have ever really advertised this exact fact but placing templates in a context may be most convenient.
https://circleci.com/docs/2.0/contexts/

Here is how you can utilize templates from environment variables: https://github.com/CircleCI-Public/slack-orb/wiki/Dynamic-Templates

KyleTryon:
Closing PR:
 - Use case should be solved by custom templates
 - Reducing lines needed by custom templates is mitigated by contexts
 - Parameters can not be used universally and would need to be tailored to work with the specific templates included in the orb
 - New parameters could not be made to work with custom or new templates.
 - Parameters are currently non-functional in current form.
 
 1. Create [custom templates](https://github.com/CircleCI-Public/slack-orb/wiki/Dynamic-Templates)
 2. Optionally you may save templates as environment variables via [contexts ](https://circleci.com/docs/2.0/contexts/)or project level environment variables. Contexts are shared for all projects.
   a. How to use templates saved to environment variables: https://github.com/CircleCI-Public/slack-orb/wiki/Dynamic-Templates

PR #254, opened 2021-03-24
[semver:patch][WEB-119] Update description to mention platform compatibility

lokst:
Update the orb description to mention Arm64 V8 compatibility


KyleTryon:
I wonder do we need to also mention the other platforms it is available for, since Windows is not supported for isntance.

This supports Linux x86_64, MacOS, and Arm64 V8

tjmcs:
👍

lokst:
> I wonder do we need to also mention the other platforms it is available for, since Windows is not supported for instance.

> This supports Linux x86_64, MacOS, and Arm64 V8

That's a great idea, I just updated the description accordingly

KyleTryon (approved):

Issue #258, opened 2021-04-08
Make scripts compatible with "bash -o nounset"


brandon-leapyear:
## Orb version:

4.3.3

## What happened:

```
+ ORB_TEST_ENV=bats-core
+ '[' /bin/bash = /bin/bash ']'
+ CheckEnvVars
+ '[' -z ****************************************************** ']'
+ '[' -z **** ']'
environment: line 116: SLACK_WEBHOOK: unbound variable
```

## Expected behavior:

Running the slack orb with `set -u` turned on for all bash shells should work

PR #259, opened 2021-04-08
Make nounset compatible [semver:minor]

brandon-leapyear:
Fixes https://github.com/CircleCI-Public/slack-orb/issues/258

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -3,12 +3,12 @@ BuildMessageBody() {
     #   If sending message, default to custom template,
     #   if none is supplied, check for a pre-selected template value.
     #   If none, error.
-    if [ -n "$SLACK_PARAM_CUSTOM" ]; then
+    if [ -n "${SLACK_PARAM_CUSTOM:-}" ]; then
         ModifyCustomTemplate
         # shellcheck disable=SC2016
         CUSTOM_BODY_MODIFIED=$(echo "$CUSTOM_BODY_MODIFIED" | sed 's/\\/\\\\/g' | sed 's/"/\\"/g' | sed 's/`/\\`/g')
         T2=$(eval echo \""$CUSTOM_BODY_MODIFIED"\")
-    elif [ -n "$SLACK_PARAM_TEMPLATE" ]; then
+    elif [ -n "${SLACK_PARAM_TEMPLATE:-}" ]; then
         TEMPLATE="\$$SLACK_PARAM_TEMPLATE"
         T1=$(eval echo "$TEMPLATE" | sed 's/"/\\"/g')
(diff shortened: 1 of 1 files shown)


src/tests/notify.bats
```diff
@@ -36,7 +36,7 @@ setup() {
     SLACK_PARAM_CUSTOM=$(cat $BATS_TEST_DIRNAME/sampleCustomTemplateWithLink.json)
     SLACK_DEFAULT_CHANNEL="xyz"
     BuildMessageBody
-    EXPECTED=$(echo "{ \"blocks\": [ { \"type\": \"section\", \"text\": { \"type\": \"mrkdwn\", \"text\": \"Sample link using environment variable in markdown <${TESTLINKURL}|LINK >\" } } ], \"text\": \"\", \"channel\": \"$SLACK_DEFAULT_CHANNEL\" }" | jq)
+    EXPECTED=$(echo "{ \"blocks\": [ { \"type\": \"section\", \"text\": { \"type\": \"mrkdwn\", \"text\": \"Sample link using environment variable in markdown <${TESTLINKURL}|LINK >\" } } ], \"text\": \"\", \"channel\": \"$SLACK_PARAM_CHANNEL\" }" | jq)
```
KyleTryon, line 39: I do not believe these should be changed in the test. The default channel value is what is mocked above. What was the reason for the name change here?

brandon-leapyear: I guess I was confused why it's using `SLACK_DEFAULT_CHANNEL` instead of `SLACK_PARAM_CHANNEL`.

https://github.com/CircleCI-Public/slack-orb/pull/259/files#diff-ef6c7403db52e40f6117d31e16d05d41ad923c6f2673e0a04cbcf97c62433c27L113-R116

Why is `SLACK_PARAM_CHANNEL` checked, but the message mentions `SLACK_DEFAULT_CHANNEL`?

KyleTryon: The value is inferred. If you do not set the `channel` parameter, the value `$SLACK_DEFAULT_CHANNEL` will be used. 
https://github.com/CircleCI-Public/slack-orb/blob/4bf71a8c2fe5666d3adc52229ce9ab030464bab1/src/commands/notify.yml#L72

This means that `SLACK_PARAM_CHANNEL` will contain the string literal `$SLACK_DEFAULT_CHANNEL`, this is evaluated so that either a string OR an environmentvariable could be used.

When we change the default of $SLACK_DEFAULT_CHANNEL we test the entire scope of this process of supporting default channels via contexts.

brandon-leapyear: I see. Ok, I've reverted it.

KyleTryon (approved):
Thank you for your contribution 🙇

PR #266, opened 2021-04-25
removed extra '\' from  notification template

junipdewan:


georgefairbairn (approved):
Agree - leaving the additional `\` in means Block Kit does not render the new line

KyleTryon (approved):
Thank you!

PR #270, opened 2021-05-05
add event override parameter

agoloAbbas:
CircleCi provide no way to run steps based on dynamic condition to run jobs in a workflow. If you want to only run "slack-notify" for a specific case you need to [export an environment variable and use it in next steps](https://circleci.com/docs/2.0/env-vars/#example-configuration-of-environment-variables). This PR allow user to pass an environment variable to trigger the slack notify and override current event parameter check.

edit: without this PR there is no way to send slack notification other than "always" , or when job fail or succeed.

### DIFF
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -20,6 +20,11 @@ parameters:
     type: enum
     enum: ["fail", "pass", "always"]
     default: "always"
+  override_event:
+    description: |
+      Send message if environment variable is not empty and override event parameter.
+    type: string
+    default: ""
   branch_pattern:
     description: |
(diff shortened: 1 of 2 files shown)


gmemstr:
@agoloAbbas Would you be able to add a test for this functionality?

KyleTryon:
@agoloAbbas Could you share more information about your specific use case? When would you use this functionality?

EricRibeiro:
@agoloAbbas, I'll close this PR since it's stale for a while now. If you still need this change, please respond to the queries above, and I'll be happy to re-open it.

Issue #272, opened 2021-05-28
Defining branch_pattern sends notifications from tag workflows


brandon-leapyear:
## Orb version:

4.3.3

## What happened:

we have a step defined like:
```yaml
- slack/notify:
    branch_pattern: master
    event: fail
    template: basic_fail_1
    channel: nightly-builds-notifications
```
where we want ONLY workflows kicked off by running the master branch to send notifications. What happened was we had a workflow kicked off from a git tag and the notification got sent from that workflow.

It seems like this behavior comes from the first line in `FilterBy`:
```bash
if [[ -z "$1" ]] || [[ -z "$2" ]];
  return
fi
```

## Expected behavior:

If I define a `branch_pattern`, it should ONLY send the notification if the workflow was kicked off by a branch AND the branch matches the pattern. Likewise, if I define a `tag_pattern`, it should ONLY send the notification if the workflow was kicked off by a git tag AND the tag matches the pattern.

I'd expect behavior more like
```diff
 FilterBy() {
-    if [ -z "$1" ] || [ -z "$2" ]; then
-      return
-    fi
 
     ...
 }
 
 ShouldPost() {
     if [ "$CCI_STATUS" = "$SLACK_PARAM_EVENT" ] || [ "$SLACK_PARAM_EVENT" = "always" ]; then
         # In the event the Slack notification would be sent, first ensure it is allowed to trigger
         # on this branch or this tag.
-        FilterBy "$SLACK_PARAM_BRANCHPATTERN" "${CIRCLE_BRANCH:-}"
-        FilterBy "$SLACK_PARAM_TAGPATTERN" "${CIRCLE_TAG:-}"
+        if [[ -n "${CIRCLE_BRANCH:-}" ]]; then
+            FilterBy "$SLACK_PARAM_BRANCHPATTERN" "$CIRCLE_BRANCH"
+        elif [[ -n "${CIRCLE_TAG:-}" ]]; then
+            FilterBy "$SLACK_PARAM_TAGPATTERN" "$CIRCLE_TAG"
+        else
+            # deduplicate logic in else block
+            exit 0
+        fi
 
         echo "Posting Status"
     else
         ...
     fi
 }
```

bobbyrenwick:
This is happening to us but in reverse, we're not setting a `branch_pattern`, just a `tag_pattern` and that's causing notifications when the build is triggered by a branch

woylie:
[...]

PR #273, opened 2021-05-28
Fail if CI env var is set but pattern is not

brandon-leapyear:
Resolves https://github.com/CircleCI-Public/slack-orb/issues/272

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -90,9 +90,7 @@ FilterBy() {
         fi
     done
     if [ "$FLAG_MATCHES_FILTER" = "false" ]; then
-        AbortPost \
-            "Current reference \"$2\" does not match any matching parameter" \
-            "Current matching pattern: $1"
+        return 1
     fi
 }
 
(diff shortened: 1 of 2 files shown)


gmemstr:
@brandon-leapyear Looks like there are some test failures to resolve. I do agree that the current behaviour should change to be more in line with what's expected from those parameters.

src/tests/notify.bats
```diff
@@ -55,52 +58,104 @@ setup() {
...
     unset SLACK_PARAM_BRANCHPATTERN
     CIRCLE_BRANCH="master"
-    run FilterBy "$SLACK_PARAM_BRANCHPATTERN" "$CIRCLE_BRANCH"
+    run ShouldPost
     echo "Error output debug: $output"
+    [[ "$output" =~ "NO SLACK ALERT" ]]
     [ "$status" -eq 0 ] # In any case, this should return a 0 exit as to not block a build/deployment.
 }
 
-@test "12: FilterBy - CIRCLE_BRANCH is empty" {
```
brandon-leapyear, line 102: the test numbers here are actually redundant; `bats` automatically prepends each test with the number, so currently the test output looks like
```
12 12: Filter By - CIRCLE_BRANCH is empty
```
Since I'm adding tests here, I figured I might as well start deduplicating the test numbers, but I can add them back for now, if you want

brandon-leapyear (commented):
@gmemstr I wasn't quite sure the best way to update the tests, but I think I figured out a decent approach. Basically changed the `FilterBy` tests to only check status code and then added `ShouldPost` tests that check for the `NO SLACK ALERT` abort message

gmemstr:
@brandon-leapyear A few more test failures :) We really appreciate you having a go at this problem.

brandon-leapyear:
try that new commit?

KyleTryon:
Retriggering tests

KyleTryon:
Thank you @brandon-leapyear 
Still showing some errors in the bats tests related to the changes.
You can run jobs locally with `circleci local execute --job bats/test`

brandon-leapyear:
Small typo in tests, that should fix it

src/scripts/notify.sh
```diff
@@ -121,22 +112,45 @@ CheckEnvVars() {
...
+            AbortPost \
+                "Branch pattern does not match: ${SLACK_PARAM_BRANCHPATTERN}" \
+                "CI was triggered by branch: ${CIRCLE_BRANCH}"
+        fi
+    elif [ -n "${CIRCLE_TAG:-}" ]; then
+        if ! FilterBy "$SLACK_PARAM_TAGPATTERN" "$CIRCLE_TAG"; then
+            AbortPost \
+                "Tag pattern does not match: ${SLACK_PARAM_TAGPATTERN}" \
+                "CI was triggered by tag: ${CIRCLE_TAG}"
+        fi
```
rwilcox, line 146: Hi my Bash isn't what it once was, I'm just checking on something because I'm seeing it in my own runs.

Does your code handle the situation where there is a tag pattern but no Git tag (aka it's running from a branch). And if so, that means, since it'll fail the `tag_pattern` that no Slack posting will happen in that instance?

brandon-leapyear: It should; the `[ -n "${CIRCLE_TAG:-}" ]` conditional will be `true` only if `CIRCLE_TAG` is set

KyleTryon:
Hey folks, very late here, apologies. I'd like to get this issue resolved. I attempted to update the branch but apparently there are still some conflicts that are not showing here in the UI, preventing me from executing the tests. If you'd be interested to try and rebase the PR, i'd be happy to try any get this merged.
I'm also happy to take this and try to re-implement the fix in a new PR.

EricRibeiro:
See: https://github.com/CircleCI-Public/slack-orb/issues/272#issuecomment-1317769296

PR #277, opened 2021-05-31
[semver:minor] #276 bugfix: added escaping of \\ and `

### LINKED ISSUE #276: Using backticks in the template causes an "invalid_arguments" error
## What happened:
When using backticks in the template (e.g. for markdown messages), they are not escaped (when providing an inline custom message, they are).
This causes the evaluation to fail (with a `-bash: content-inside-backticks: command not found`, which in turn leaved the rendered template (`T2`) empty, getting the following response from Slack:
```json
{"ok":false,"error":"invalid_arguments","response_metadata":{"messages":["[ERROR] missing required field: channel"]}}
```

## Expected behavior:
Escaping the backticks too.

## Additional Information:
`\\` is also escpaed for inline custom messages but not for templates.

OrBin:
I believe this PR should close #276.

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -10,6 +10,7 @@ BuildMessageBody() {
         T2=$(eval echo \""$CUSTOM_BODY_MODIFIED"\")
     elif [ -n "${SLACK_PARAM_TEMPLATE:-}" ]; then
         TEMPLATE="\$$SLACK_PARAM_TEMPLATE"
+        # shellcheck disable=SC2016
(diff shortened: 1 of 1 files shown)


gmemstr (changes requested):
@OrBin Some failing tests to take look at :)

OrBin:
@gmemstr I think it should be fixed now. Do you know why the tests are not being rerun?

OrBin:
@gmemstr The checks pass now

gmemstr:
On second thought, I'm marking this as a minor since I'm not sure if anyone is relying on the current behaviour. Otherwise, looks good :) Thank you!

gmemstr (approved):

Issue #285, opened 2021-06-28
Add "mentions" to the basic_success_1 template.


ozomer:
"Mentions" is useful not only in fail cases, but also in success cases.

Another option is to create a different `basic_success_2` template, or `basic_success_with_mentions_1` template.

GarvitSinghal47:
ozomer, i will like to work on it can you plz explain some more detail about it

HennaAbbas:
@ozomer you can use the mentions parameter with the `basic_success_1` template. What slack orb version are you using? Can you try using the latest version: 4.5.2 ?

Additionally, you can use the [block kit builder](https://app.slack.com/block-kit-builder/) to create your own custom template.

HennaAbbas:
The functionality was added in the following PR:  [Add mentions to basic_success_1.json #286](https://github.com/CircleCI-Public/slack-orb/pull/286).

PR #286, opened 2021-06-28
Add mentions to basic_success_1.json

ozomer:
* Add mentions to basic_success_1 template, for a similar usecase
like in the basic_fail_1 template.

See: https://github.com/CircleCI-Public/slack-orb/issues/285


KyleTryon:
Thank you @ozomer. I think we can get away with calling this a patch as no functionality was actually changed.  Merging now.

KyleTryon (approved):

PR #291, opened 2021-08-16
Check for SLACK_WEBHOOK before SLACK_ACCESS_TOKEN [semver:patch]

Peeja:
`SLACK_WEBHOOK` is no longer supported; instead, `SLACK_ACCESS_TOKEN` and `SLACK_PARAM_CHANNEL` are used. We've been checking for those first, and bailing if they're not found. But people upgrading to 4.x won't understand the error they see about a missing `SLACK_ACCESS_TOKEN`.

Instead, check for and warn about `SLACK_WEBHOOK` first to make it clear why more setup is required.

(Source: Tried to upgrade, was confused. 😉 )

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -104,6 +104,11 @@ FilterBy() {
 }
 
 CheckEnvVars() {
+    if [ -n "${SLACK_WEBHOOK:-}" ]; then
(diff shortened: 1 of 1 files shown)


KyleTryon (approved):
Much appreciated 🙇

PR #296, opened 2021-09-14
Update slack orb examples documentation

m0ab:
These commits will update the documentation for the examples as used by the Slack Orb to update how the node orb version is declared and update the versions of the Slack and Node orb themselves to be accurate against the current Orb versions.

### DIFF
--- a/src/examples/custom_notification.yml
+++ b/src/examples/custom_notification.yml
@@ -6,7 +6,7 @@ description: |
 usage:
   version: 2.1
   orbs:
-    slack: circleci/slack@4.1
+    slack: circleci/slack@4.4.4
   jobs:
     notify:
       docker:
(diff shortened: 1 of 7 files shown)


KyleTryon:
@m0ab Thank you for this 🙇 . Could I ask you to leave the patch version digit off of the semver examples? The patch will update too frequently and these will become out of date fairly quickly. If we can pin to a fairly stable minor version though, hopefully it would go without needing update for some time.

m0ab:
@KyleTryon Sure thing, no problem. Will fix this later today.

m0ab:
@KyleTryon I removed the patch version from the node and slack orb docs. Let me know if that looks good.

KyleTryon:
Thank you @m0ab. Apologies but I am sweeping up some maintenance and this is going to rapidly become out of date. I will make the changes to the docs at the end of this little maintenance sweep. Thank you for bringing attention to this.

PR #303, opened 2021-10-26
[semver:patch] fix: update basic success newline escaping

### LINKED ISSUE #255: Cannot get circleci/slack@4.1 to Work. invalid_arguments error
## Orb version:

This does not work
```
- slack/notify:
     event: pass
     template: basic_success_1
- slack/notify:
     event: fail
     template: basic_fail_1
```

I've tried setting the `channel` property above or a `SLACK_DEFAULT_CHANNEL` environment variable in the context. 

## What happened:

```
Posting Status
Sending to Slack Channel: *********
{
  "ok": false,
  "error": "invalid_arguments"
}
```

## Expected behavior:

Notification should be sent.

## Additional Information:

It would be helpful to add logging to help debug this.

iggy:
The basic_success_1 template doesn't escape newlines properly. Update it
to match the others (basically `s/\n/\\n/`). I see it was like this at
one point, but then the `\\n` were changed to `\n` at some point. I'm
guessing since we don't have problems with the other templates and they
all use the `\\n` format, that it should be using `\\n`.

maybe
Fixes #255

### DIFF
--- a/src/message_templates/basic_success_1.json
+++ b/src/message_templates/basic_success_1.json
@@ -23,19 +23,19 @@
 			"fields": [
 				{
 					"type": "mrkdwn",
-					"text": "*Project*:\n $CIRCLE_PROJECT_REPONAME"
+					"text": "*Project*:\\n $CIRCLE_PROJECT_REPONAME"
 				},
 				{
 					"type": "mrkdwn",
-					"text": "*Branch*:\n $CIRCLE_BRANCH"
-                },
-                {
-					"type": "mrkdwn",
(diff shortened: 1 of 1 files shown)


wyardley:
I'm seeing this as well -- would be great if someone could approve / merge this one.

gmemstr (approved):

wyardley:
@gmemstr @iggy I'm still seeing lots of extra newlines in output of various fields from various templates, even with the latest version. Will try to open a case

PR #311, opened 2021-12-02
(RE-1507) Add explicit $BASH_ENV execution

kvimber:
When calling slack notify via the orb or in a command, using `$BASH_ENV` doesn't seem to work. Adding this explicit call lets others continue to customize the command via `$BASH_ENV` and have those customizations change the slack behavior.

Testing Results (links internal to CircleCI org):

- [job](https://app.circleci.com/pipelines/github/circleci/server/21570/workflows/40b18d91-30b6-4b4b-bfbd-555542335cc1/jobs/75332) (defined in [this branch](https://github.com/circleci/server/compare/main...repro/RE-1507)) showing the `$SLACK_PARAM_MENTIONS` being set in `$BASH_ENV`, but not being read by the `slack/notify` command correctly for the [slack post]()
- [job](https://app.circleci.com/pipelines/github/circleci/server/21821/workflows/3c823a3c-e5d9-447a-814d-a7ba035d1cf7/jobs/76872) showing this fix being passed the `$SLACK_PARAM_MENTIONS` & printing the value after running the `$BASH_ENV` explicitly

Please let me know if there's anything I need or missed for this PR, I'm having trouble finding a contributor doc or guides.

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -144,6 +144,7 @@ ShouldPost() {
 # This is done so this script may be tested.
 ORB_TEST_ENV="bats-core"
 if [ "${0#*$ORB_TEST_ENV}" = "$0" ]; then
+    # shellcheck disable=SC1090
     . $BASH_ENV
     CheckEnvVars
     . "/tmp/SLACK_JOB_STATUS"
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -104,6 +104,17 @@ FilterBy() {
     fi
 }
 
+SetupEnvVars() {
+    echo "BASH_ENV file: $BASH_ENV"
+    if [ -f "$BASH_ENV" ]; then
+        echo "Exists. Sourcing into ENV"
+        # shellcheck disable=SC1090
+        . $BASH_ENV
+    else
+        echo "Does Not Exist. Skipping file execution"
+    fi
+}
+
 CheckEnvVars() {
     if [ -n "${SLACK_WEBHOOK:-}" ]; then
         echo "It appears you have a Slack Webhook token present in this job."
@@ -144,8 +155,7 @@ ShouldPost() {
 # This is done so this script may be tested.
 ORB_TEST_ENV="bats-core"
 if [ "${0#*$ORB_TEST_ENV}" = "$0" ]; then
-    # shellcheck disable=SC1090
-    . $BASH_ENV
+    SetupEnvVars
     CheckEnvVars
     . "/tmp/SLACK_JOB_STATUS"
     ShouldPost
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -144,6 +144,7 @@ ShouldPost() {
 # This is done so this script may be tested.
 ORB_TEST_ENV="bats-core"
 if [ "${0#*$ORB_TEST_ENV}" = "$0" ]; then
+    . $BASH_ENV
     CheckEnvVars
     . "/tmp/SLACK_JOB_STATUS"
     ShouldPost


KyleTryon (approved):
Certainly can't hurt! Good call thank you! I am triggering the tests on our end manually now.

src/scripts/notify.sh
```diff
@@ -144,6 +144,7 @@ ShouldPost() {
 # This is done so this script may be tested.
 ORB_TEST_ENV="bats-core"
 if [ "${0#*$ORB_TEST_ENV}" = "$0" ]; then
+    . $BASH_ENV
```
KyleTryon, line 147: Add `# shellcheck disable=SC1090` to the line above to ignore this error.
https://github.com/koalaman/shellcheck/wiki/SC1090

We are safe to ignore this, the variable is known within CircleCI.

KyleTryon (changes requested):
Thank you @kvimber, could we ask you to add a comment above this line to ignore an erroneous error?

kvimber:
of course! Added in 21f6e2b

KyleTryon:
Wel... I was not expecting this, and I feel like it should be immediately obvious, I may need a moment to figure out what failed here. For some reason, we failed to load $BASH_ENV in each test, though we did clearly properly find the environment variable, we thought it was empty for some reason. Going to need to dig in here a little.

kvimber:
Thanks for figuring this out @KyleTryon & making the update super easy for me (in 9924e28)! I thought I was just getting to the point where it's big enough to warrant its own function (& I like that the name lines up perfectly with `CheckEnvVars` too), so I did that. I'm on the fence though, size-wise, so let me know if you'd rather not have it that way & I'll make it so.

I also have a bit of a preference to be verbose since CircleCI has such great folding support, you can see an example of this running in [this server step](https://app.circleci.com/pipelines/github/circleci/server/22181/workflows/4e8495c8-64fd-4897-929b-2d7c0157ebf0/jobs/78864?invite=true#step-115-169) leading to [this slack post](https://circleci.slack.com/archives/C02N8LD1LV9/p1638833904000200). Let me know if you think of better messaging here! I really appreciate your quick eyes on these, thanks for all your help!

KyleTryon:
@kvimber , @HennaAbbas and I are looking at the orb and wanted to clarify the use case here as we might have a more simple alternative. Here, you are modifying the "internal" environment variables of the orb, in order to take advantage of the existing templates. 

The alternative we were thinking about, what if instead of modifying the "internals" of the orb, you instead saved what ever value you wanted to an environment variable `MY_MENTIONS` and instead, passed this value to the orb like so:

```
- slack/notify:
          branch_pattern: production
          event: fail
          mentions: $MY_MENTIONS
          template: basic_fail_1
```

Now the above example will not work today (i believe) because the `mentions` parameter is taking in a string and I believe would print the literal environment variable name rather than its value. However, with a smaller change to the orb, we could use parameter expansion on `$SLACK_PARAM_MENTIONS` to get the "true" value, just as we do with some of our other orbs when we need to expand an environment variable: https://github.com/CircleCI-Public/aws-eks-orb/blob/0e679e924af0c0dfaf48d6cf0a5d9acc451b4995/src/scripts/create-cluster.sh#L2

What do you think?

kvimber:
Hi @KyleTryon & @HennaAbbas. I think I must be misunderstanding the option you're describing. Do we need the new variable `MY_MENTIONS`?

I tried to recreate this by passing the `mentions` param (in [this server code](https://github.com/circleci/server/compare/0093a72...cd4602f)) with the `SLACK_PARAM_MENTIONS` var & then updated [my custom slack-orb](https://app.circleci.com/pipelines/github/circleci/server/22434/workflows/6306c59c-a041-4b04-9f27-0601d92d7e37/jobs/80366?invite=true#step-115-2) to run that line you mentioned & ignore this PR code. Doing that, you can see in the output that the env var isn't set and that shows on the [corresponding slack notification](https://circleci.slack.com/archives/C02N8LD1LV9/p1639007715000300).

KyleTryon:
> Hi @KyleTryon & @HennaAbbas. I think I must be misunderstanding the option you're describing. Do we need the new variable `MY_MENTIONS`?
> 
> I tried to recreate this by passing the `mentions` param (in [this server code](https://github.com/circleci/server/compare/0093a72...cd4602f)) with the `SLACK_PARAM_MENTIONS` var & then updated [my custom slack-orb](https://app.circleci.com/pipelines/github/circleci/server/22434/workflows/6306c59c-a041-4b04-9f27-0601d92d7e37/jobs/80366?invite=true#step-115-2) to run that line you mentioned & ignore this PR code. Doing that, you can see in the output that the env var isn't set and that shows on the [corresponding slack notification](https://circleci.slack.com/archives/C02N8LD1LV9/p1639007715000300).

Sorry for the delay.

Yes that sounds right. My thinking is, within the context of your job, perhaps you set an environment variable `MY_ENV`, this will need to be added to `BASH_ENV`. This is very similar to how Dynamic Templates are suggested here: https://github.com/CircleCI-Public/slack-orb/wiki/Dynamic-Templates.

We would just need to ensure that environment variables were evaluated so they do not post the literal string `$MY_ENV`

kvimber:
No worries Kyle, thanks for the reply! Looking at the Dynamic Templates doc made me think that perhaps I was just including the variable wrong. [Fixing that](https://github.com/circleci/server/commit/c25e6137987f9862ec6a5ec70f47c7127e3e8c6a) doesn't seem to have helped though.

I'm including the `SLACK_PARAM_MENTIONS` variable in the `BASH_ENV` as we've discussed and at the above "my custom slack-orb" link, you can see I'm executing the line to evaluate the environment variable, but sadly [that doesn't work](https://app.circleci.com/pipelines/github/circleci/server/23108/workflows/2ad17be7-e044-4cb1-a5fd-d8f3b6755917/jobs/84233?invite=true#step-115-170) ([slack notification](https://circleci.slack.com/archives/C02N8LD1LV9/p1639518739000100)). So either I must be misunderstanding something, or the approach you're suggesting doesn't work. I'd be happy to pair on this with you to show what's going on in the code so we can figure out which of these it is.

EricRibeiro:
Thank you @kvimber!

Awesome stuff you have going here 👏. I am going to take this off Kyle's hand and have a look at what might be the issue.

EricRibeiro (approved):
Thank you!

PR #317, opened 2022-01-12
Feat: Add environment variable support to the "mentions" parameter [semver:minor]

EricRibeiro:
As [proposed](https://github.com/CircleCI-Public/slack-orb/pull/311#issuecomment-988205044) in the PR #311, this change adds support for an environment variable in the [mentions](https://github.com/CircleCI-Public/slack-orb/blob/master/src/commands/notify.yml#L33) parameter.

Acceptable arguments now are:

- Environment variable
```
      - slack/notify:
          template: basic_success_1
          mentions: "$MY_MENTION"
          event: always

```

- Direct mention:
```
- slack/notify:
          template: basic_fail_1
          mentions: "@orbs"
          event: always
```

- Empty
```
      - slack/notify:
          template: basic_success_1
          event: always
```

Since we need a way to differentiate direct mentions from env variables, the env variable parameter has to contain the "$" (dollar sign).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,12 +62,20 @@ jobs:
           name: Verify SLACK_JOB_STATUS is written
           command: |
             grep "pass" /tmp/SLACK_JOB_STATUS
+      - run:
+          name: Export mention variable
+          command: |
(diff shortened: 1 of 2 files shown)


EricRibeiro:
#322 is looking better than this one, so I'm closing it to avoid confusion.

PR #324, opened 2022-01-24
[semver:patch] Sudo fix

brivu:


.circleci/config.yml
```diff
@@ -102,7 +102,7 @@ workflows:
       - orb-tools/lint
       - shellcheck/check:
           dir: ./src/scripts
-          exclude: SC2148,SC2001,SC1091,SC2086,SC2116,SC2046,SC2005,SC2002
+          exclude: SC2148,SC2001,SC1091,SC2086,SC2116,SC2046,SC2005,SC2002,SC2039
```
KyleTryon, line 105: ^ Why was this added

EricRibeiro: This line was added to support the operator `&>`. But, as we can't compromise sh POSIX compliance, I went back on this change and used an alternate operator.

EricRibeiro (approved):
All good!

Issue #300, opened 2021-09-29
Preventing debian repo error by passing `--allow-releaseinfo-change` to apt-get


tzimmermann:
I would like to suggest a more general solution to the issue https://github.com/CircleCI-Public/slack-orb/issues/292.

We are using the slack/notify orb for a couple of different jobs with different executors.
Some of these are running as root (for example [this one from cypress](https://github.com/cypress-io/cypress-docker-images/blob/master/browsers/node14.15.0-chrome86-ff82/Dockerfile)) and some are not.

Currently, there is no easy way to get around the error in https://github.com/CircleCI-Public/slack-orb/issues/292 except for running `apt-get update --allow-releaseinfo-change` with and without `sudo` prefix in several places.

What about simply adding `--allow-releaseinfo-change` to the `apt-get update` command inside the notify script here?
That should solve our issues and prevent anyone else hitting this in the future and needing to work around it.

https://github.com/CircleCI-Public/slack-orb/blob/9650da83d92512c1a0e8be50e7482a6a7f6bbcca/src/scripts/notify.sh#L71

EricRibeiro:
Thanks for opening this issue, @tzimmermann.

You have a good point. Another solution is to curl jq from their GitHub repository.

I'll run a few tests to check which strategy makes more sense and open a PR shortly.

PR #327, opened 2022-01-28
Fix: Change jq source to avoid errors with package managers [semver:minor]

### LINKED ISSUE #294: JQ is not always installed in docker images
## Orb version:

4.4.x

## What happened: 

used a non CircleCI Docker image ( amazon/aws-cli ) Jq is not always installed 

## Expected behavior: 

The orb should check if JQ is installed. If it is not, install jq 

jq: command not found

EricRibeiro:
Closes #294 and #300.

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -72,7 +72,7 @@ ModifyCustomTemplate() {
 }
 
 InstallJq() {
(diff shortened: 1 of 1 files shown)


KyleTryon (approved):
Great 👍

PR #328, opened 2022-01-31
Update notify.sh to add a option to override SUDO variable [semver:patch]

ganezasan:
We've been using the slack orb in our team repository, and noticed an issue that the notify script failed with `sudo: no password was provided` error. This is because, the root user has configured the password. 

Build URL: https://app.circleci.com/pipelines/github/circleci/mongo-config-management/3287/workflows/3be9fc63-2298-4d9c-82ff-3a5418851a25/jobs/5206

We want to avoid using `sudo` on the runner executor, so I've changed the notify script to override the SUDO variable by the environment variable. Also, added some small fixes.

## Changes
- remove `export`. There is no reason to use export here. In shell script, a variable is in the global scope as default.
- override SUDO if there is a `SUDO` environment variable
- remove unnecessary spaces

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -1,8 +1,8 @@
 #!/bin/sh
 if ! which sudo > /dev/null 2>&1 || [ "$(id -u)" = 0 ]; then
-    SUDO="${SUDO:-''}";
+    SUDO=${SUDO:-""};
 else
-    SUDO="${SUDO:-sudo}";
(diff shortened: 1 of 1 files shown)


KyleTryon (dismissed):
approve for [semver:patch], apologies for the inconvenience caused by the change. Thank you for your assistance.

KyleTryon (approved):
merging. Thank you!

PR #329, opened 2022-01-31
[semver:minor] Add skip functionality to slack

katebygrace:
### SEMVER Update Type: 
Minor 

### Description: 
Add a "skip" option
### Motivation: 
We have a need to dynamically set whether or not the slack orb runs that would be picked during the run (so parameters would not work in this use case). This pull request adds a "skip" option, so we can then set "event" to be something we can pick during the CI run. 
### Closes Issues: 
None 

ISSUE URL
Checklist:
 All new jobs, commands, executors, parameters have descriptions.
No new commands added
 Usage Example version numbers have been updated. 
I don't think that was needed for this PR, but could be wrong?
 Changelog has been updated.
I didn't see a changelog in this repository. Am i missing something?

### DIFF
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -19,9 +19,9 @@ parameters:
     default: ""
   event:
     description: |
-      In what event should this message send? Options: ["fail", "pass", "always"]
+      In what event should this message send? Options: ["fail", "pass", "skip", "always"]
     type: enum
-    enum: ["fail", "pass", "always"]
+    enum: ["fail", "pass", "skip", "always"]
     default: "always"
   branch_pattern:
     description: |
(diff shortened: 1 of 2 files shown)


EricRibeiro:
Hi @katebygrace 👋 

Thanks for contributing to our repo! Could you give me an example of your use case?

Also, have you tried [logic statements](https://circleci.com/docs/2.0/configuration-reference/#logic-statement-examples)? I can't say for sure without more context, but they might do the trick for you.

katebygrace:
Yes - that's what I meant when i said "parameters don't work" but I can elaborate. We have a script that's not terribly dissimilar to the semantic version publish that most orbs use (including this one) - so we have a script that decides what to do if its a major / minor / patch / skip during the run and publishes and does some other steps accordingly. Skip still "succeeds" (I started making skip fail, but I really do not want it to fail there! Fail is for real CI failures) and we were doing the "pass" event, but need to skip slack sometimes, but this is decided _during_ the script. The examples of using "when" that I saw key off of parameters set at runtime. So I'd then set the `event` to be a variable that I'd set while running the semantic version script, after publish is complete.

katebygrace:
Any more info needed here?

PR #333, opened 2022-02-07
Feat: Add template inference based on job status if no template is provided [semver:minor]

### LINKED ISSUE #306: single notify for success / failure?
## Describe Request:
IIRC, with v 3.x of this orb, you could use the same notify for both success / failure (optionally with different messages), and without specifying a template. With the new orb, you have to use something like:

```yaml
            - slack/notify:
                event: pass
                template: basic_success_1
            - slack/notify:
                event: fail
                template: basic_fail_1
```

IMO, it would be nice to have a basic notify command (or a fallback when no template / event is passed) such that you could just have
```yaml
 - slack/notify
```

and get a basic success template on success or basic fail template on failure.

## Supporting Documentation Links:

EricRibeiro:
This PR closes #306 by allowing a single notify command for successful and failed pipeline scenarios. 

If the job has `pass` status, the template used will be `basic_success_1`. On the other hand, if it has `fail`, the template `basic_fail_1` will be used instead.

With this change, users can now do:

> \- slack/notify

And receive notifications for both cases.


src/commands/notify.yml
```diff
@@ -9,6 +9,7 @@ parameters:
       Select which template to use for the notification by its name. The name must be available as an environment variable.
       The built-in templates can be found and previewed at: https://github.com/CircleCI-Public/slack-orb/wiki#templates.
       Alternatively, you can create and use your own dynamic templates: https://github.com/CircleCI-Public/slack-orb/wiki/Dynamic-Templates.
+      If left empty and no custom template is provided, the template will be automatically selected based on the pipeline status.
```
KyleTryon, line 12: Status of the _job_ not the pipeline

src/examples/notify_with_inferred_template.yml
```diff
@@ -0,0 +1,26 @@
+
```
KyleTryon, line 1: I recommend not including this example. The name isn't obvious.  and the example is long but can not really demonstrate the outcome.

KyleTryon (changes requested):

src/examples/notify_with_inferred_template.yml
EricRibeiro: Should I include an example that makes more sense or no example at all?

KyleTryon (approved):

wyardley:
Thanks @EricRibeiro! I think this will be useful!

PR #336, opened 2022-02-15
Clean up log files at script exit to avoid buildup of leftover log fi…

atodd-circleci:
Clean up log files at script exit to avoid buildup of leftover log files on persistent runners.

Make variable use safer by quoting all values.

I have tested the trap behavior using `dash`, which is the POSIX-compliant shell that ships with Ubuntu. I think that the `ash` used in Alpine Linux should be functionally equivalent. Of course, it also works in `bash`.

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -3,6 +3,25 @@ LOG_PATH=$(mktemp -d)
 JQ_PATH=/usr/local/bin/jq
 POST_TO_SLACK_LOG=post-to-slack.json
 
+stdmsg() {
+    local IFS=' '
+    printf '%s\n' "$*"
+}
+
+TrapExit() {
+  # It is critical that the first line capture the exit code. Nothing else can come before this.
+  # The exit code recorded here comes from the command that caused the script to exit.
+  local exit_status="$?"
+
+  rm -rf "$LOG_PATH"
+
+  if [ "$exit_status" -ne 0 ]; then
+    stdmsg 'The script did not complete successfully.'
+    stdmsg 'The exit code was '"$exit_status"
+  fi
+}
+trap TrapExit EXIT
+
 BuildMessageBody() {
     # Send message
     #   If sending message, default to custom template,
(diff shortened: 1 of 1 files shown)


src/scripts/notify.sh
```diff
@@ -3,6 +3,25 @@ LOG_PATH=$(mktemp -d)
 JQ_PATH=/usr/local/bin/jq
 POST_TO_SLACK_LOG=post-to-slack.json
 
+stdmsg() {
```
atodd-circleci, line 6: `echo` should be avoided in shell scripting because its behavior is inconsistent across implementations; see [this Stack Exchange question](https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo) for more details.

KyleTryon:
@atodd-circleci,

The value in the log files is that they will be available in the container on CircleCI cloud when you SSH into the box for debugging. If the files are removed automatically you will not be able to use them for debugging.

atodd-circleci:
@KyleTryon The contents of the log file are already printed to stdout by `tee`; SSHing to the completed job wouldn't add any more value:

https://github.com/CircleCI-Public/slack-orb/pull/336/files#diff-ef6c7403db52e40f6117d31e16d05d41ad923c6f2673e0a04cbcf97c62433c27R68

Further, if `SLACK_PARAM_DEBUG` is set, the same information is written to the console in a non-JSON form.

It looks to me like one of these is redundant.

I can't find anything [in the documentation](https://circleci.com/developer/orbs/orb/circleci/slack) specifying how the log is supposed to work, so I can't judge what is happening here, because I don't understand what the requirement is.

It looks like, prior to the introduction of `mktemp`, when this was using the same log file name for all runs, if the same container was used for multiple runs of this Orb, then the output of all of the runs would be appended to the same file, I think. Introducing `mktemp` broke that.

I'm not sure what the intent here is, but I think the full output is available in the build logs right now. Without a clear definition of the expected behavior and why that behavior is useful, I don't know how to proceed.

atodd-circleci:
Closed in favor of https://github.com/CircleCI-Public/slack-orb/pull/337

PR #337, opened 2022-02-18
Restrict logs creation to when the debug flag is set to true [semver:patch]

EricRibeiro:
With this change, the `notify` command will only save the request payload and response if the `debug` parameter is set to `true`.

This PR also modifies how we handle log creation. Before, we were appending new entries to a single log file. Now we are keeping them more concise and simple to handle by splitting requests and payload into different files and generating new directories for each `slack/notify` command.

### DIFF
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -59,7 +59,7 @@ parameters:
       description: |
        View payload and response being sent to Slack API.
        Enable to view full payload being sent to Slack and response being received from the API call.
-       Unredacted content can be viewed by re-running the job with SSH and accessing the content of the log files referenced in the job output.
+       Redacted content can be viewed by re-running the job with SSH and accessing the log files referenced in the job output.
       type: boolean
       default: false
 steps:
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -57,8 +57,9 @@ parameters:
       default: true
   debug:
       description: |
-       View payload and response being sent to slack api.
-       Enable to view full payload being sent to slack and response being recieved from the API call.
+       View payload and response being sent to Slack API.
+       Enable to view full payload being sent to Slack and response being received from the API call.
+       Unredacted content can be viewed by re-running the job with SSH and accessing the content of the log files referenced in the job output.
       type: boolean
       default: false
 steps:
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -57,7 +57,7 @@ parameters:
       default: true
   debug:
       description: |
-       View payload and response being sent to Slack API.
+       View payload and response being sent to the Slack API.
        Enable to view full payload being sent to Slack and response being received from the API call.
        Redacted content can be viewed by re-running the job with SSH and accessing the log files referenced in the job output.
       type: boolean
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -57,9 +57,9 @@ parameters:
       default: true
   debug:
       description: |
-       View payload and response being sent to the Slack API.
        Enable to view full payload being sent to Slack and response being received from the API call.
        Redacted content can be viewed by re-running the job with SSH and accessing the log files referenced in the job output.
+       When run in a persistent build environment such as CircleCI Runner, these debug log files may remain in the system's temporary filesystem indefinitely and accumulate over time.
       type: boolean
       default: false
 steps:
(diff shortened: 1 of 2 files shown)


KyleTryon (dismissed):

src/scripts/notify.sh
```diff
@@ -42,14 +40,18 @@ PostToSlack() {
         echo "Sending to Slack Channel: $i"
         SLACK_MSG_BODY=$(echo "$SLACK_MSG_BODY" | jq --arg channel "$i" '.channel = $channel')
         if [ -n "${SLACK_PARAM_DEBUG:-}" ]; then
-            echo "The message body being sent to Slack is: $SLACK_MSG_BODY"
+            printf "%s" "$SLACK_MSG_BODY" > "$SLACK_MSG_BODY_LOG"
```
atodd-circleci, line 43: I'd encourage using a `\n` here or using [the approach described here in "Getting non-clobbered output from command substitution" (which is more general, if more confusing)](https://www.etalabs.net/sh_tricks.html) to guarantee that the file contains a valid POSIX text file, which always has a trailing newline.

`\n` is more straightforward, though, and probably not harmful in this situation.
```suggestion
            printf "%s\n" "$SLACK_MSG_BODY" > "$SLACK_MSG_BODY_LOG"
```

src/scripts/notify.sh
```diff
@@ -42,14 +40,18 @@ PostToSlack() {
...
+            echo "The message body being sent to Slack can be found below. To view redacted values, rerun the job with SSH and access: ${SLACK_MSG_BODY_LOG}"
+            echo "$SLACK_MSG_BODY"
         fi
         SLACK_SENT_RESPONSE=$(curl -s -f -X POST -H 'Content-type: application/json' -H "Authorization: Bearer $SLACK_ACCESS_TOKEN" --data "$SLACK_MSG_BODY" https://slack.com/api/chat.postMessage)
-        cat $LOG_PATH/$POST_TO_SLACK_LOG | jq --argjson message "$SLACK_MSG_BODY"  --argjson response "$SLACK_SENT_RESPONSE" \
-            '. += [{"slackMessageBody": $message, "slackSentResponse": $response}]' | tee $LOG_PATH/$POST_TO_SLACK_LOG
+        
         if [ -n "${SLACK_PARAM_DEBUG:-}" ]; then
-            echo "The response from the API call to slack is : $SLACK_SENT_RESPONSE"
+            printf "%s" "$SLACK_SENT_RESPONSE" > "$SLACK_SENT_RESPONSE_LOG"
```
atodd-circleci, line 50:
```suggestion
            printf "%s\n" "$SLACK_SENT_RESPONSE" > "$SLACK_SENT_RESPONSE_LOG"
```

src/commands/notify.yml
```diff
@@ -57,8 +57,9 @@ parameters:
       default: true
   debug:
       description: |
-       View payload and response being sent to slack api.
-       Enable to view full payload being sent to slack and response being recieved from the API call.
+       View payload and response being sent to Slack API.
+       Enable to view full payload being sent to Slack and response being received from the API call.
+       Redacted content can be viewed by re-running the job with SSH and accessing the log files referenced in the job output.
```
atodd-circleci, line 62: Thank you for making grammar improvements, this makes the documentation much easier to read.

I'm not sure if it is in line with our other documentation, but it might be worth adding something about persistent environments.

```suggestion
       Redacted content can be viewed by re-running the job with SSH and accessing the log files referenced in the job output.
       When run in a persistent build environment such as CircleCI Runner, these debug log files may remain in the system's temporary filesystem indefinitely and accumulate over time.
```

src/commands/notify.yml
```diff
@@ -57,8 +57,9 @@ parameters:
       default: true
   debug:
       description: |
-       View payload and response being sent to slack api.
-       Enable to view full payload being sent to slack and response being recieved from the API call.
+       View payload and response being sent to Slack API.
```
atodd-circleci, line 60:
```suggestion
       View payload and response being sent to the Slack API.
```

atodd-circleci (approved):
Thank you -- this creates an Orb that can be used across a wide variety of environments.

I would be curious to know where the temporary directories are being created on systems that require root to create directories in `/tmp`.

atodd-circleci (approved):

KyleTryon (approved):

PR #347, opened 2022-04-19
feat: passthrough circleci_host in on-hold job

acburdine:
- while the circleci_host parameter was added to the notify command, it isn't being passed through from the on-hold job


EricRibeiro (approved):
Thank you for your contribution! ✌️

PR #354, opened 2022-06-08
feat: [semver:patch] Add mentions to success_tagged_deploy_1.json

gritzkoo:
* Add mentions to success_tagged_deploy_1.json template, for a similar use-case

like in the basic_fail_1 template

### DIFF
--- a/src/message_templates/success_tagged_deploy_1.json
+++ b/src/message_templates/success_tagged_deploy_1.json
@@ -32,6 +32,15 @@
 			}
 		},
 		{
+			"type": "section",
+			"fields": [
+				{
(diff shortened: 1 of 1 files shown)


EricRibeiro:
Olá, @gritzkoo 👋 

Thank you for your contribution! While I appreciate the idea, I don't believe we should introduce a new section to an existing template as it will change its behaviour for everyone currently using it.

Did you consider using a [custom template](https://github.com/CircleCI-Public/slack-orb/wiki#custom-message-template)?

gritzkoo:
Thank you for the advice! I'll close this PR.

PR #364, opened 2022-09-20
Fix invert_match to work with multiple branch patterns

edennis:
First off - thanks for adding the `invert_match` param in the last release. This makes dealing with notifications for unknown branch names much easier. Unfortunately, in its current state it only works if you have a single pattern in the `branch_pattern` pattern. This is due to the "match fast" logic with the `break` when a match occurs. When inverting this logic, we need to ensure that ALL of the patterns don't match instead of ANY.

Example:
```yaml
branch_pattern: 'foo,bar,baz'
invert_match: true
```
If we're currently deploying the "baz" branch, "foo" does not match, so with `invert_match` true, `FLAG_MATCHES_FILTER` gets set to true, the `for`-loop is exited and the slack notification is incorrectly sent.

This PR simplifies the logic by only taking into account whether or not `invert_match` is true and then negates the `FLAG_MATCHES_FILTER` behavior when deciding whether or not to exit.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ workflows:
           requires:
             [orb-tools/lint, orb-tools/review, orb-tools/pack, shellcheck/check]
           # Use a context to hold your publishing token.
-          context: orb-publishing
+          context: orb-publisher
           filters:
             tags:
               only: /.*/
--- a/.circleci/test-deploy.yml
+++ b/.circleci/test-deploy.yml
@@ -114,7 +114,7 @@ workflows:
             - orb-tools/review
             - orb-tools/pack
             - integration-test-templates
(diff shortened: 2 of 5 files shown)


EricRibeiro (dismissed):
Thank you for your contribution 🙇

EricRibeiro (approved):

edennis:
Thanks for merging @EricRibeiro! When can we expect a version bump?

bear:
💛 approved

EricRibeiro:
> Thanks for merging @EricRibeiro! When can we expect a version bump?

Sorry for not getting back to you sooner @edennis. This change was released in [v4.11.0](https://circleci.com/developer/orbs/orb/circleci/slack?version=4.11.0).

PR #368, opened 2022-11-02
fix: invert match conditionals and bats tests

EricRibeiro:
## Changes

- Group conditionals and add `\` to properly break the line.
- Add the `bats` orb to run unit tests.
- Replace `==` with `=~` in the test 13.

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -92,7 +92,6 @@ InstallJq() {
 }
 
 FilterBy() {
-    set -x
     if [ -z "$1" ] || [ -z "$2" ]; then
       return
     fi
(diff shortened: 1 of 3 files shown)


src/scripts/notify.sh
```diff
@@ -105,8 +105,8 @@ FilterBy() {
         fi
     done
     # If the invert_match parameter is set, invert the match.
-    if [ "$FLAG_MATCHES_FILTER" = "false" ] && [ "$SLACK_PARAM_INVERT_MATCH" -eq 0 ] ||
-       [ "$FLAG_MATCHES_FILTER" = "true" ] && [ "$SLACK_PARAM_INVERT_MATCH" -eq 1 ]
+    if { [ "$FLAG_MATCHES_FILTER" = "false" ] && [ "$SLACK_PARAM_INVERT_MATCH" -eq 0 ]; } || \
```
Jaryt, line 108: This could probably be done better. I believe "$FLAG_MATCHES_FILTER" is a number before it gets quoted, if so, something like this could be performed:
 
(psuedo code)

```
match = $FLAG_MATCHES_FILTER

if [ invert match ]; then
    match = $(1 - $FLAG_MATCHES_FILTER) # invert match boolean
fi
```

Jaryt (approved):
Approving since this is code that was already there. However, something like my comment may be better than what existed before.

bear:
💛 approved

PR #369, opened 2022-11-02
feat: exit if host OS is Windows

### LINKED ISSUE #358: Inform user of non-support for Windows environment
## Orb version:

circleci/slack@4.10.1

## What happened:

Not exactly a bug, but I was thinking it may be good for the `notify` command to fail early if the detected OS is Windows.
Given that we currently do not support Windows in this Orb (see #282), it may be nice to fail with a useful message early then.

EricRibeiro:
## Changes

- Prevent job from running if host OS is Windows (closes #358)

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -175,10 +175,20 @@ SetupLogs() {
     fi
 }
 
+ExitIfWindows() {
+    os="$(uname -s | tr '[:upper:]' '[:lower:]')"
+    if printf '%s\n' "$os" | grep --quiet 'msys*\|cygwin*' ; then
+        printf '%s\n' "Windows is not supported by this orb."
+        printf '%s\n' "For more information, see: https://github.com/CircleCI-Public/slack-orb/wiki/FAQ."
+        exit 1
+    fi
+}
+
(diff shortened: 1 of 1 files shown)


src/scripts/notify.sh
```diff
@@ -175,10 +175,20 @@ SetupLogs() {
...
+        printf '%s\n' "For more information, see: https://github.com/CircleCI-Public/slack-orb/wiki/FAQ."
+        exit 1
+    fi
+}
+
 # Will not run if sourced from another script.
 # This is done so this script may be tested.
 ORB_TEST_ENV="bats-core"
 if [ "${0#*"$ORB_TEST_ENV"}" = "$0" ]; then
+    ExitOnWindows
```
Jaryt, line 191: nit pick: `ExitOnWindows` is a slightly confusing name. Might be better along the lines of `CheckOnWindows` or `ExitIfWindows`. I say this since the check of windows is inside the called function.

Jaryt (approved):
Approving with comment

Jaryt (approved):
Thank you!

Jaryt (approved):

bear:
💛 approved

PR #373, opened 2022-11-03
fix: replace `--quiet` with `-q` in grep to support Alpine

EricRibeiro:
## Changes

- Replace grep's flag to support Alpine.


Jaryt (approved):
good ol' alpine

bear:
💛 approved

PR #376, opened 2022-11-16
fix: escape backslashes

### LINKED ISSUE #279: Using dynamic template with env variables does not work
## Orb version:
4.4.2

## What happened:
I'm trying to send a list of commits in a Slack message but so far, without success, encountering this error `parse error: Invalid numeric literal at line 1`. From the documentation, it is a bit unclear to me how to use dynamic templates that are using env variables so any hint would be more than welcomed!

This is what I have done so far:

- `config.yml`:
```
  steps:
      - checkout
      - jq/install
      - run:
          name: Export commits list
          command: |
            echo 'export COMMITS_LIST=$(git log --oneline origin/master...origin/develop | jq -aRs .)' >> $BASH_ENV
            echo 'export SLACK_AUTO_DEPLOY_TEMPLATE=$(cat .circleci/slack/auto-deploy.json)' >> $BASH_ENV
            source $BASH_ENV
      - slack/notify:
          event: always
          channel: 'XXX'
          template: SLACK_AUTO_DEPLOY_TEMPLATE
```

- `.circleci/slack/auto-deploy.json`:
```
{
  "blocks":[
    {
      "type":"section",
      "text":{
        "type":"mrkdwn",
        "text":"Latest changes: \`\`\`${COMMITS_LIST}\`\`\`"
      }
    }
  ]
}
```

## Expected behavior:

Using dynamic template with an environment variable should work without raising a "parse error".

## Additional Information:

- I've tried piping to `jq -aRs .` to help but no luck so far 😔
- Might be related to #264 

Thanks a lot for your help!

EricRibeiro:
# Motivation

Closes #279 

# Changes

- Add backslashes to the list of characters to be sanitized in environment variables.
- Remove `git` dependency from integration tests since Alpine doesn't have it.
- Improve sanitization unit tests by adding "Expected vs Actual" output in case of failure.

### DIFF
--- a/.circleci/test-deploy.yml
+++ b/.circleci/test-deploy.yml
@@ -82,7 +82,7 @@ jobs:
       - run:
           name: Dynamically populate the mention and export the template as an environment variable
           command: |
-            echo 'export COMMIT_MESSAGE="$(git log --format=oneline -n 1 $CIRCLE_SHA1)"' >> $BASH_ENV
+            echo 'export COMMIT_MESSAGE="This text comes from an environment variable"' >> $BASH_ENV
             echo 'export SLACK_PARAM_MENTIONS="$COMMIT_MESSAGE"' >> $BASH_ENV
             echo 'export MY_ORB_TEMPLATE=$(cat src/message_templates/basic_success_1.json)' >> $BASH_ENV
       - slack/notify:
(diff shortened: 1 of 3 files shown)


KyleTryon (approved):
Great work

PR #448, opened 2024-03-25
fix: re-running pipelines

atanass:
Description:

This is a fix for the latest thread message posting capability.

Motivation:

There is a misbehaviour when there is a Slack message posted but the pipeline fails. In these occasions, if a user initiates a  "Rerun from failed" and the failed job actually passes and the pipeline gets to the slack job, intended to post a message in a thread, the message will be posted as a new one.
This is because of the way the cache key is formed (`{{ .Environment.CIRCLE_WORKFLOW_ID }}`) which is different when rerunning the pipeline.
This PR adds another cache key to be relied on in such scenarios. It will now be falling back to using the git revision, which will be the same with every rerun of the pipeline.

Checklist:
[x] All new jobs, commands, executors, parameters have descriptions.
[ ] Usage Example version numbers have been updated.
[ ] Changelog has been updated.

### DIFF
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -103,7 +103,8 @@ steps:
       steps:
         - restore_cache:
             keys:
-              - cache-<< parameters.thread_id >>-{{ .Environment.CIRCLE_WORKFLOW_ID }}
+              - cache-{{ .Revision }}-{{ .Environment.CIRCLE_WORKFLOW_ID }}
+              - cache-{{ .Revision }}
   - run:
       when: always
       name: << parameters.step_name >>
@@ -134,6 +135,10 @@ steps:
           equal: [ "", <<parameters.thread_id>> ]
       steps:
         - save_cache:
-            key: cache-<< parameters.thread_id >>-{{ .Environment.CIRCLE_WORKFLOW_ID }}
+            key: cache-{{ .Revision }}-{{ .Environment.CIRCLE_WORKFLOW_ID }}
+            paths:
+              - /tmp/SLACK_THREAD_INFO
+        - save_cache:
+            key: cache-{{ .Revision }}
             paths:
               - /tmp/SLACK_THREAD_INFO
--- a/src/commands/notify.yml
+++ b/src/commands/notify.yml
@@ -103,8 +103,7 @@ steps:
       steps:
         - restore_cache:
             keys:
-              - cache-{{ .Revision }}-{{ .Environment.CIRCLE_WORKFLOW_ID }}
-              - cache-{{ .Revision }}
+              - cache-{{ .Environment.CIRCLE_PIPELINE_ID }}
   - run:
       when: always
       name: << parameters.step_name >>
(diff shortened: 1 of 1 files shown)


atanass:
- Before:
    - Rerun from failed: starts a new thread
    - Rerun from start: starts a new thread
- Now:
    - Rerun from failed: posts a reply to initial message
    - Rerun from start: posts a reply to initial message

atanass:
@otremblay @aaronstillwell

src/commands/notify.yml
```diff
@@ -134,6 +135,10 @@ steps:
...
       steps:
         - save_cache:
-            key: cache-<< parameters.thread_id >>-{{ .Environment.CIRCLE_WORKFLOW_ID }}
+            key: cache-{{ .Revision }}-{{ .Environment.CIRCLE_WORKFLOW_ID }}
+            paths:
+              - /tmp/SLACK_THREAD_INFO
+        - save_cache:
+            key: cache-{{ .Revision }}
             paths:
               - /tmp/SLACK_THREAD_INFO
```
aaronstillwell, line 144: I think just the first `save_cache` is required

atanass: I see your point, but the goal here is to search for an explicit fallback cache. If I remove the second `save_cache` and change the second key in `restore_cache` by adding a `-` at the end, the rerun pipeline won't get to the fallback cache key because there won't be such, because `{{ .Environment.CIRCLE_WORKFLOW_ID }}` is never going to be empty, and `cache-{{ .Revision }}-` will never be initiated.

aaronstillwell: @atanass our docs do not seem to be as clear as they really should be on the matter, so I'll address that separately, but keys are used in a sort of prefix search.

So by creating the key `cache-{{ .Revision }}-{{ .Environment.CIRCLE_WORKFLOW_ID }}` we are doing enough already, since a prefix search using `cache-{{ .Revision }}` will find the _most recent_ matching key with that prefix.

Does that make sense?

If so, I'll raise a separate issue over on the docs to update https://circleci.com/docs/caching/#restoring-cache

aaronstillwell: Actually maybe my eyesight is the problem rather than the docs, as that page does mention it albeit it's subtle `Each cache key is namespaced to the project and retrieval is prefix-matched`

atanass: I see your point. I think you're right. Let me test that and will update accordingly.
Thank you.

aaronstillwell: We also _just_ made an update that could make this even simpler, as we now have the environment variable `CIRCLE_PIPELINE_ID` available. This ID is the same between re-runs, and could replace the revision/workflow ID combo.

aaronstillwell: Note I think this was literally just added today, so technically would not work for any pipelines created before today.

atanass: Brilliant. That's exactly what I needed! Pushing an updated version.

otremblay (approved):

orb-publisher:
Your orb has been published to the CircleCI Orb Registry.
You can view your published orb on the CircleCI Orb Registry at the following link: 
https://circleci.com/developer/orbs/orb/circleci/slack?version=4.13.2

PR #471, opened 2024-12-18
Fix "warning: regexp escape sequence `\"' is not a known regexp operator"

y-yagi:
Since v5.0.1, awk doesn't treat `\"` as a regexp operator. So the current implementation causes the following warning when using awk >= v5.0.1.

```
awk: cmd. line:1: warning: regexp escape sequence `\"' is not a known regexp operator
```

This PR replaces `\"` to `"` to fix the warning.

Ref: https://github.com/KittyKatt/screenFetch/issues/627

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -323,11 +323,7 @@ SanitizeVars() {
         # Escape newlines.
         sanitized_value="$(printf '%s' "$sanitized_value" | tr -d '\r' | awk 'NR > 1 { printf("\\n") } { printf("%s", $0) }')"
         # Escape double quotes.
-        if [ "$PLATFORM" = "windows" ]; then
-            sanitized_value="$(printf '%s' "$sanitized_value" | awk '{gsub(/"/, "\\\""); print $0}')"
-        else
-            sanitized_value="$(printf '%s' "$sanitized_value" | awk '{gsub(/\"/, "\\\""); print $0}')"
-        fi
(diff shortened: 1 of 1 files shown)


marboledacci:
What environments have you seen this warning on?

y-yagi:
Our environment is Ubuntu 22.04.1 LTS. Awk is GNU Awk 5.1.0.

```bash
$ awk --version
GNU Awk 5.1.0, API: 3.0 (GNU MPFR 4.1.0, GNU MP 6.2.1)
$ echo '"abc"' | awk '{gsub(/\"/, "\\\""); print $0}'
awk: cmd. line:1: warning: regexp escape sequence `\"' is not a known regexp operator
```

marboledacci:
I replicated the behavior, and I'm seeing the warning but I see the replacement still works so I'm not sure if the change is justified as in most environments the warning doesn't happen.
For the environments I tested, the substitution works fine with both versions, but I'm not sure if it could break for some other environments. The change depends a lot on the awk implementation which is usually very different across environments.
Do you have any more context about this that justifies more the change? I don't want to risk breaking the orb in an unexpected environment just to remove a harmless warning.

y-yagi:
I assume this warning came from https://git.savannah.gnu.org/cgit/gawk.git/commit/?id=3ddc932b.

From my understanding, the old behavior was just meaningless. So the above change has shown the warning for users to notice it. So the behavior hasn't changed at all.

But, this is about GNU awk.

I'm not sure about the supported environment of this orb, but it's difficult to guarantee all environments. 
So I will fix this issue for our environment with other approaches and close this PR. Thanks!

PR #466, opened 2024-10-22
Add a retry option to the notify command

marboledacci:
I'm adding a retry to the notify command as requested in #426

PR #476, opened 2025-03-30
Add option to update the contents of a notification instead of posting a threaded reply to it

christophe-SA:
**SEMVER Update Type:**
- [ ] Major
- [x] Minor
- [ ] Patch

## Description:
add `update_notification` flag to `notify` job, which will update the top level message in the thread instead of appending a new message to the bottom of the thread. If no thread is found or no thread is specified, defaults to a regular notification posting.

also added a charset to the slack api curl statement to address the `missing charset` warning from the slack api

## Motivation:
By updating the top level message instead of posting threaded replied, it is easy to update the progress of a job without sending new notifications for each update. This also allows for links or actions to be "enabled" after the job completes, such as when a deploy is successful, providing next steps and additional info in the top level message. 

not dissimilar to https://github.com/CircleCI-Public/slack-orb/pull/466/files
 **Closes Issues:**
-  ISSUE URL

## Checklist:

- [x] All new jobs, commands, executors, parameters have descriptions.
- [ ] Usage Example version numbers have been updated.
- [ ] Changelog has been updated.

### DIFF
--- a/src/scripts/notify.sh
+++ b/src/scripts/notify.sh
@@ -59,7 +59,7 @@ NotifyWithRetries() {
     local success_request=false
     local retry_count=0
     while [ "$retry_count" -le "$SLACK_PARAM_RETRIES" ]; do
-        if SLACK_SENT_RESPONSE=$(curl -s -f -X POST -H 'Content-type: application/json' -H "Authorization: Bearer $SLACK_ACCESS_TOKEN" --data "$SLACK_MSG_BODY" "$1"); then
+        if SLACK_SENT_RESPONSE=$(curl -s -f -X POST -H 'Content-type: application/json; charset=utf-8' -H "Authorization: Bearer $SLACK_ACCESS_TOKEN" --data "$SLACK_MSG_BODY" "$1"); then
             echo "Notification sent"
             success_request=true
             break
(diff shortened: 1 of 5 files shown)


src/scripts/notify.sh
```diff
@@ -59,7 +59,7 @@ NotifyWithRetries() {
     local success_request=false
     local retry_count=0
     while [ "$retry_count" -le "$SLACK_PARAM_RETRIES" ]; do
-        if SLACK_SENT_RESPONSE=$(curl -s -f -X POST -H 'Content-type: application/json' -H "Authorization: Bearer $SLACK_ACCESS_TOKEN" --data "$SLACK_MSG_BODY" "$1"); then
+        if SLACK_SENT_RESPONSE=$(curl -s -f -X POST -H 'Content-type: application/json; charset=utf-8' -H "Authorization: Bearer $SLACK_ACCESS_TOKEN" --data "$SLACK_MSG_BODY" "$1"); then
```
marboledacci, line 62: Why is this required?

christophe-SA: the slack api returns a warning about the missing charset. 
see https://api.slack.com/methods/chat.postMessage#warnings

marboledacci:
I have run the tests for this, and they are passing, but I don't think it is working as expected. @christophe-SA is the test I added a good example of what this PR supports? How should the result of it be, because I'm getting a normal slack message from it.

PR #477, opened 2025-03-30
Docs for dynamically generating notify message template

christophe-SA:
added documentation for dynamically generating a template from a script and then passing it properly into the slack/notify job

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,26 @@ The Slack Orb comes with a number of included templates to get your started with
         }
   ```
 
+  ## Dynamically Generating or Loading Importing a Custom Message
+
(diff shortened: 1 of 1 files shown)


marboledacci:
I'm not familiar with that behavior. I will have to do some tests to validate that it actually works as you say and then I can approve it.

christophe-SA:
Thanks. I'm running a fork with this and my update message pr for my org's circle configs. I'm dumping the stdout of my script into a temp file and loading that into the bash env. I found the technique from googling somewhere, but I'm having trouble finding the original link again. lmk if you run into any issues - i'd love to switch back from my fork at some point

marboledacci (approved):

PR #496, opened 2025-12-17
fix: check for curl and add it if needed

alex-mustapha:


david-montano-circleci:
Hello @alex-mustapha . Thank you for creating this PR. I see this tries to merge against an older version of the main branch. There are some conflicts to resolve.
I would also say that if there are packages to be installed then they must be added not inside this orb logic but as steps that run before invoking this logic so I would advise against this curl installation here.

Issue #502, opened 2026-04-20
Add example: single notification for parallel test jobs


jlw-shimizu-hrtk:
## Describe Request:

When using `parallelism`, `slack/notify` fires once per node, resulting in
duplicate notifications for a single test run. It would be helpful to have
an example showing how to aggregate parallel job results and send exactly
one notification based on the combined outcome.

## Examples:

```yaml
jobs:
  notify:
    docker:
      - image: cimg/base:stable
    steps:
      - run:
          name: Check parallel test results
          command: |
            JOBS=$(curl -sf \
              -H "Circle-Token: $CIRCLECI_TOKEN" \
              "https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/job")

            FAILED=$(echo "$JOBS" | jq '[
              .items[] | select(.name == "test" and (.status == "failed" or .status == "failing"))
            ] | length')

            if [ "$FAILED" -gt "0" ]; then
              echo "$FAILED node(s) failed."
              exit 1
            fi
            echo "All nodes passed."

      - slack/notify:
          event: fail
          template: basic_fail_1
      - slack/notify:
          event: pass
          template: basic_success_1

workflows:
  main:
    jobs:
      - test:
          parallelism: 6
      - notify:
          requires:
            - test: [success, failed]
```

Happy to open a PR if this fits the direction of the project.

## Supporting Documentation Links:

- CircleCI API v2 – Get a workflow's jobs: https://circleci.com/docs/api/v2/index.html#operation/listWorkflowJobs
- `requires` with job status filters: https://circleci.com/docs/configuration-reference/#requires

HennaAbbas:
Hi @jlw-shimizu-hrtk , you can use a  special job to handle this use case that ignores the status/outcome of an upstream job: 
https://circleci.com/docs/guides/orchestrate/workflows/#ignore-the-outcome-of-an-upstream-job

Alternatively, we are in preview with the new CircleCI Slack Integration: 
https://circleci.com/docs/guides/integration/slack-integration/

jlw-shimizu-hrtk:
Thanks for the feedback, @HennaAbbas 

I intentionally used `[success, failed]` rather than `terminal` in the example to make the intent more explicit for readers who may not be familiar with `terminal` yet — but happy to swap that if you prefer consistency.

[...]

PR #503, opened 2026-05-19
docs: add example aggregating parallel test results into one notification

jlw-shimizu-hrtk:
## Summary

Adds a new example, `src/examples/notify_aggregated_parallel_results.yml`, showing how to send **exactly one** Slack notification for a parallel (`parallelism`) test job — instead of one per node.

The example uses a dedicated `notify` job that:
1. Runs after the parallel `test` job using `requires: - test: [success, failed]` (so it executes regardless of pass/fail).
2. Queries the CircleCI API v2 (`/workflow/{id}/job`) to collect every parallel node's status.
3. Exits non-zero if any node failed, then uses `slack/notify` (`event: fail` / `event: pass`) to send a single aggregated notification.

This addresses a UX problem raised in #502: with `parallelism`, `slack/notify` currently fires once per node, producing duplicate notifications for a single logical test run. There was no documented pattern for collapsing those into one.

## Context

- Issue: #502
- Maintainer go-ahead: https://github.com/CircleCI-Public/slack-orb/issues/502#issuecomment-4480080108 (\"please feel free to open a PR\")
- The pattern itself is in production use in our own CircleCI pipelines, so the YAML reflects a working configuration rather than a theoretical sketch.

## Notes for reviewers

- Uses `$CIRCLE_TOKEN` (CircleCI's standard env var name) rather than `CIRCLECI_TOKEN` used in the original issue body.
- File naming follows the existing `src/examples/` convention (snake_case, verb-led).
- No changes to `README.md` — existing examples are surfaced via the Orb Registry rather than direct README links, so this PR keeps to that convention. Happy to add a README pointer in a follow-up if preferred.

## Test plan

- [x] Confirm YAML lints under the repo's existing config (`.yamllint`).
- [x] Confirm the example renders correctly in the Orb Registry / docs pipeline.

### DIFF
--- a/src/examples/notify_aggregated_parallel_results.yml
+++ b/src/examples/notify_aggregated_parallel_results.yml
@@ -3,8 +3,9 @@ description: |
   duplicate notifications for a single logical test run.
   This example shows how to aggregate the results of parallel jobs into a single Slack
   notification by adding a dedicated "notify" job that:
-    1. Runs only after all parallel "test" nodes finish (using `requires` with `[success, failed]`
-       so it runs whether the parallel run passed or failed).
+    1. Runs only after all parallel "test" nodes reach a terminal state (using `requires`
+       with `terminal` so it runs whether the parallel run passed, failed, was canceled,
+       or timed out).
     2. Queries the CircleCI API v2 for the statuses of all "test" job nodes.
     3. Sends exactly one Slack notification reflecting the combined outcome.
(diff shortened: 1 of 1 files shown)


jlw-shimizu-hrtk:
Friendly ping — this has been open since May with no review yet.

Happy to rebase or adjust if anything needs changing. If the example isn't a direction you want in the README, feel free to close it; no hard feelings either way.

